PR #235, opened 2019-05-09
Azure pipelines

ingeknudsen:


azure-pipelines/build-and-push-pipeline.yml
```diff
@@ -0,0 +1,47 @@
+# Docker image
+# Build a Docker image to deploy, run, or push to a container registry.
+# Add steps that use Docker Compose, tag images, push to a registry, run an image, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/docker
```
nemzes: Should remove this generic comment or, even better, replace with a short description of the purpose of this pipeline.

azure-pipelines/build-only-pipeline.yml
```diff
@@ -0,0 +1,35 @@
+# Docker image
+# Build a Docker image to deploy, run, or push to a container registry.
+# Add steps that use Docker Compose, tag images, push to a registry, run an image, and more:
+# https://docs.microsoft.com/azure/devops/pipelines/languages/docker
```
nemzes: As above

azure-pipelines/build-and-push-pipeline.yml
ingeknudsen: Have provided a new description

azure-pipelines/build-only-pipeline.yml
ingeknudsen: Have provided a new description

PR #237, opened 2019-05-13
RA-147  LimitRange - Request bigger then default limit - pod not created

nemzes:
## Container resources

Resolves the conflict by assigning a `limit` equal to the `request` value if `request` is larger than `default limit` and no `limit` has been provided.

The `GetDefault[CPU|Memory][Limit|Request]` functions were refactored to be reusable.

### DIFF
--- a/pkg/apis/deployment/kubedeployment.go
+++ b/pkg/apis/deployment/kubedeployment.go
@@ -3,6 +3,7 @@ package deployment
 import (
 	"strings"
 
+	"github.com/equinor/radix-operator/pkg/apis/defaults"
 	"github.com/equinor/radix-operator/pkg/apis/kube"
 	v1 "github.com/equinor/radix-operator/pkg/apis/radix/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -165,14 +166,44 @@ func getSecurityContextForContainer() *corev1.SecurityContext {
 }
 
 func getResourceRequirements(deployComponent v1.RadixDeployComponent) *corev1.ResourceRequirements {
+
+	defaultLimits := map[corev1.ResourceName]resource.Quantity{
+		corev1.ResourceName("cpu"):    *defaults.GetDefaultCPULimit(),
+		corev1.ResourceName("memory"): *defaults.GetDefaultMemoryLimit(),
+	}
+
 	// if you only set limit, it will use the same values for request
 	limits := corev1.ResourceList{}
 	requests := corev1.ResourceList{}
+
 	for name, limit := range deployComponent.Resources.Limits {
-		limits[corev1.ResourceName(name)], _ = resource.ParseQuantity(limit)
+		resName := corev1.ResourceName(name)
+
+		if limit != "" {
+			limits[resName], _ = resource.ParseQuantity(limit)
+		}
+
+		// TODO: We probably should check some hard limit that cannot by exceeded here
 	}
+
 	for name, req := range deployComponent.Resources.Requests {
-		requests[corev1.ResourceName(name)], _ = resource.ParseQuantity(req)
+		resName := corev1.ResourceName(name)
+
+		if req != "" {
+			requests[resName], _ = resource.ParseQuantity(req)
+
+			if _, hasLimit := limits[resName]; !hasLimit {
+				// There is no defined limit, but there is a request
+				reqQuantity := requests[resName]
+				if reqQuantity.Cmp(defaultLimits[resName]) == 1 {
+					// Requested quantity is larger than the default limit
+					// We use the requested value as the limit
+					limits[resName] = requests[resName].DeepCopy()
+
+					// TODO: If we introduce a hard limit, that should not be exceeded here
+				}
+			}
+		}
 	}
 
 	if len(limits) <= 0 && len(requests) <= 0 {
--- a/pkg/apis/deployment/kubedeployment_test.go
+++ b/pkg/apis/deployment/kubedeployment_test.go
@@ -1,0 +1,113 @@
+package deployment
+
+import (
+	"os"
+	"testing"
+
+	"github.com/equinor/radix-operator/pkg/apis/defaults"
+	"github.com/equinor/radix-operator/pkg/apis/utils"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestGetResourceRequirements_Provide_Requests_Limits(t *testing.T) {
+	request := map[string]string{
+		"cpu":    "0.1",
+		"memory": "32Mi",
+	}
+
+	limit := map[string]string{
+		"cpu":    "0.5",
+		"memory": "64Mi",
+	}
+
+	component := utils.NewDeployComponentBuilder().
+		WithResource(request, limit).
+		BuildComponent()
+	requirements := getResourceRequirements(component)
+
+	assert.Equal(t, 0, requirements.Requests.Cpu().Cmp(resource.MustParse("0.1")), "CPU request should be included")
+	assert.Equal(t, 0, requirements.Requests.Memory().Cmp(resource.MustParse("32Mi")), "Memory request should be included")
+
+	assert.Equal(t, 0, requirements.Limits.Cpu().Cmp(resource.MustParse("0.5")), "CPU limit should be included")
+	assert.Equal(t, 0, requirements.Limits.Memory().Cmp(resource.MustParse("64Mi")), "Memory limit should be included")
+}
+
+func TestGetResourceRequirements_Provide_Requests_Only(t *testing.T) {
+	request := map[string]string{
+		"cpu":    "0.2",
+		"memory": "128Mi",
+	}
+
+	component := utils.NewDeployComponentBuilder().
+		WithResourceRequestsOnly(request).
+		BuildComponent()
+	requirements := getResourceRequirements(component)
+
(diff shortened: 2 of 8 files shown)


pkg/apis/deployment/kubedeployment_test.go
```diff
@@ -0,0 +1,113 @@
...
+	"os"
+	"testing"
+
+	"github.com/equinor/radix-operator/pkg/apis/defaults"
+	"github.com/equinor/radix-operator/pkg/apis/utils"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+func TestGetResourceRequirements_Provide_Requests_Limits(t *testing.T) {
```
ingeknudsen: I prefer the format Test<function>_<condition>_<expected>, but I am aware that there are as many opinions as there are developers... In this case it would be TestGetResourceRequirements_BothProvided_BothReturned

pkg/apis/deployment/kubedeployment_test.go
```diff
@@ -0,0 +1,113 @@
...
+		"cpu":    "0.5",
+		"memory": "64Mi",
+	}
+
+	component := utils.NewDeployComponentBuilder().
+		WithResource(request, limit).
+		BuildComponent()
+	requirements := getResourceRequirements(component)
+
+	assert.Equal(t, 0, requirements.Requests.Cpu().Cmp(resource.MustParse("0.1")), "CPU request should be included")
```
ingeknudsen: It took a little while before I understood the assert. I would have expected this to be either:
assert.True(t, requirements.Requests.Cpu().Cmp(resource.MustParse("0.1")), "CPU request is returned as expected")

or:
assert.Equal(t, parseQuantity("0.1"), requirements.Requests.Cpu()), "CPU request should be included")

pkg/apis/deployment/kubedeployment_test.go
```diff
@@ -0,0 +1,113 @@
...
+	assert.Equal(t, 0, requirements.Requests.Memory().Cmp(resource.MustParse("32Mi")), "Memory request should be included")
+
+	assert.Equal(t, 0, requirements.Limits.Cpu().Cmp(resource.MustParse("0.5")), "CPU limit should be included")
+	assert.Equal(t, 0, requirements.Limits.Memory().Cmp(resource.MustParse("64Mi")), "Memory limit should be included")
+}
+
+func TestGetResourceRequirements_Provide_Requests_Only(t *testing.T) {
```
ingeknudsen: Same as above

pkg/apis/deployment/kubedeployment_test.go
```diff
@@ -0,0 +1,113 @@
...
+		"cpu":    "0.2",
+		"memory": "128Mi",
+	}
+
+	component := utils.NewDeployComponentBuilder().
+		WithResourceRequestsOnly(request).
+		BuildComponent()
+	requirements := getResourceRequirements(component)
+
+	assert.Equal(t, 0, requirements.Requests.Cpu().Cmp(resource.MustParse("0.2")), "CPU request should be included")
```
ingeknudsen: Same as above

pkg/apis/deployment/kubedeployment_test.go
```diff
@@ -0,0 +1,113 @@
...
+	assert.Equal(t, 0, requirements.Requests.Memory().Cmp(resource.MustParse("128Mi")), "Memory request should be included")
+
+	assert.Equal(t, 0, requirements.Limits.Cpu().Cmp(resource.MustParse("0")), "Missing CPU limit should be 0")
```
ingeknudsen: Not quite sure I understand why missing = 0. Is it a technicality?

pkg/apis/deployment/kubedeployment_test.go
```diff
@@ -0,0 +1,113 @@
...
+	requirements := getResourceRequirements(component)
+
+	assert.Equal(t, 0, requirements.Requests.Cpu().Cmp(resource.MustParse("6")), "CPU request should be included")
+	assert.Equal(t, 0, requirements.Requests.Memory().Cmp(resource.MustParse("0")), "Missing memory request should be 0")
+
+	assert.Equal(t, 0, requirements.Limits.Cpu().Cmp(resource.MustParse("6")), "CPU limit should be same as request")
+	assert.Equal(t, 0, requirements.Limits.Memory().Cmp(resource.MustParse("0")), "Missing memory limit should be 0")
+}
+
+func TestGetResourceRequirements_Provide_Limits_Only(t *testing.T) {
```
ingeknudsen: Empty test function

pkg/apis/deployment/kubedeployment.go
```diff
@@ -165,14 +166,44 @@ func getSecurityContextForContainer() *corev1.SecurityContext {
...
+		corev1.ResourceName("memory"): *defaults.GetDefaultMemoryLimit(),
+	}
+
 	// if you only set limit, it will use the same values for request
 	limits := corev1.ResourceList{}
 	requests := corev1.ResourceList{}
+
 	for name, limit := range deployComponent.Resources.Limits {
-		limits[corev1.ResourceName(name)], _ = resource.ParseQuantity(limit)
+		resName := corev1.ResourceName(name)
```
ingeknudsen: Why exactly do you need this loop? Is it to copy the structure over from deployComponent.Resources.Limits? Why can we just make a deep copy as you do below?

pkg/apis/deployment/kubedeployment.go
```diff
@@ -165,14 +166,44 @@ func getSecurityContextForContainer() *corev1.SecurityContext {
 }
 
 func getResourceRequirements(deployComponent v1.RadixDeployComponent) *corev1.ResourceRequirements {
+
```
ingeknudsen: This function doesn't appear to be called from anywhere. Could you use my test to verify it is called correctly (deployment_test.go):

func TestObjectSynced_RequestNoLimit_LimitEqualsRequest(t *testing.T) {
    tu, client, radixclient := setupTest()

    // Test
    applyDeploymentWithSync(tu, client, radixclient,
        utils.ARadixDeployment().
            WithAppName("anyName").
            WithEnvironment("test").
            WithComponents(
                utils.NewDeployComponentBuilder().
                    WithResource(map[string]string{
                        "memory": "64Mi",
                        "cpu":    "250m",
                    }, nil)))

    envNamespace := utils.GetEnvironmentNamespace("anyName", "test")
    deployments, _ := client.ExtensionsV1beta1().Deployments(envNamespace).List(metav1.ListOptions{})
    assert.Equal(t, parseQuantity("64Mi"), deployments.Items[0].Spec.Template.Spec.Containers[0].Resources.Requests["memory"])
    assert.Equal(t, parseQuantity("250m"), deployments.Items[0].Spec.Template.Spec.Containers[0].Resources.Requests["cpu"])
    assert.Equal(t, parseQuantity("64Mi"), deployments.Items[0].Spec.Template.Spec.Containers[0].Resources.Limits["memory"])
    assert.Equal(t, parseQuantity("250m"), deployments.Items[0].Spec.Template.Spec.Containers[0].Resources.Limits["cpu"])
    })
}

pkg/apis/deployment/kubedeployment.go
nemzes: This loop was here before; I've just added a guard in case the limit isn't set so we don't set it to zero. The `limits` structure is used further down to decide whether limits have actually been declared, so having zero values in it is not helpful.

pkg/apis/deployment/kubedeployment_test.go
nemzes: Yeah, I'm not happy with it but there's no other way I can see to test equality for `Quantity` objects. `assert.True` won't work, since `Cmp()` returns `-1`, `0`, or `1`. And your suggestion to test equality with `parseQuantity("0.1")` won't work because these are objects.

pkg/apis/deployment/kubedeployment_test.go
nemzes: It's the way it is implemented in https://github.com/kubernetes/api/blob/master/core/v1/resource.go

pkg/apis/deployment/kubedeployment.go
nemzes: It's used here: https://github.com/equinor/radix-operator/blob/73ea8c3966526dd1d3c26eaf9cbd673c06c7b6c4/pkg/apis/deployment/kubedeployment.go#L118

pkg/apis/deployment/kubedeployment_test.go
ingeknudsen: One way it could be written is:
	assert.Equal(t, parseQuantity("0.1"), *requirements.Requests.Cpu(), "CPU request should be included")
	assert.Equal(t, parseQuantity("32Mi"), *requirements.Requests.Memory(), "Memory request should be included")

	assert.Equal(t, parseQuantity("0.5"), *requirements.Limits.Cpu(), "CPU limit should be included")
	assert.Equal(t, parseQuantity("64Mi"), *requirements.Limits.Memory(), "Memory limit should be included")

nemzes: Fixed for non-zero values. For the "not present is zero" we need a different assertion (fortunately `Quantity` has a `IsZero()` function), since equality of a struct is not exactly straightforward, especially via the assertion library: https://github.com/stretchr/testify/blob/master/assert/assertions.go#L56

PR #254, opened 2019-06-07
Parameterize rollingUpdate and readinessPeriod parameters to env vari…

thezultimate:
…ables

### DIFF
--- a/pkg/apis/defaults/readinessprobe.go
+++ b/pkg/apis/defaults/readinessprobe.go
@@ -1,0 +1,34 @@
+package defaults
+
+import (
+	"os"
+	"strconv"
+)
+
+// Environment variables that define default readiness probe parameters for containers
+const (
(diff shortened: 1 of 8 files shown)


pkg/apis/defaults/readinessprobe.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+	"strconv"
+)
+
+// Environment variables that define default readiness probe parameters for containers
+const (
+	OperatorReadinessProbeInitialDelaySeconds = "RADIXOPERATOR_APP_READINESS_PROBE_INITIAL_DELAY_SECONDS"
+	OperatorReadinessProbePeriodSeconds       = "RADIXOPERATOR_APP_READINESS_PROBE_PERIOD_SECONDS"
+	ReadinessProbeInitialDelaySecondsDefault  = 5
+	ReadinessProbePeriodSecondsDefault        = 10
```
nemzes: It's probably best to not have default values; this will mask error scenarios. Just let it blow up.

PR #255, opened 2019-06-11
Ra 37 external alias

ingeknudsen:
Will make it possible to add radix config like this:

  dnsExternalAlias:
    - alias: test1.iknu.stian.tech
      component: frontend
      environment: dev
    - alias: test2.iknu.stian.tech
      component: frontend
      environment: dev


pkg/apis/deployment/secrets.go
```diff
@@ -84,6 +167,18 @@ func (deploy *Deployment) garbageCollectSecretsNoLongerInSpec() error {
 	return nil
 }
 
+func (deploy *Deployment) listSecretsForComponent(component radixv1.RadixDeployComponent, labelSelector string) (*v1.SecretList, error) {
```
thezultimate: `component` variable is not used in this function. Remove it?

pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -83,6 +89,41 @@ func validateDNSAppAlias(app *radixv1.RadixApplication) []error {
 	return errs
 }
 
+func validateDNSExternalAlias(app *radixv1.RadixApplication) []error {
+	errs := []error{}
+
+	distictAlias := make(map[string]bool)
```
thezultimate: Possible typo: `distinctAlias`

pkg/apis/deployment/secrets.go
ingeknudsen: Refactored as suggested

pkg/apis/radixvalidators/validate_ra.go
ingeknudsen: Renamed, as suggested

PR #263, opened 2019-06-27
RA-376 Implement Active Cluster Fix Ingress

thezultimate:
…vironment variable, handle ingresses

### DIFF
--- a/pkg/apis/deployment/ingress.go
+++ b/pkg/apis/deployment/ingress.go
@@ -18,14 +18,13 @@ import (
 const (
 	appAliasTLSSecretName       = "app-wildcard-tls-cert"
 	clusterDefaultTLSSecretName = "cluster-wildcard-tls-cert"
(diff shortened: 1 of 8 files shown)


pkg/apis/deployment/ingress.go
```diff
@@ -18,14 +18,13 @@ import (
 const (
 	appAliasTLSSecretName       = "app-wildcard-tls-cert"
 	clusterDefaultTLSSecretName = "cluster-wildcard-tls-cert"
+	activeClusterTLSSecretName  = "active-cluster-wildcard-tls-cert"
 )
 
 func (deploy *Deployment) createIngress(deployComponent v1.RadixDeployComponent) error {
 	namespace := deploy.radixDeployment.Namespace
 	clustername, err := deploy.kubeutil.GetClusterName()
```
keaaa: could GetClusterName ever fail? if yes - check for error, if not - use _ instead of instantiating err variable.

PR #278, opened 2019-08-05
Add tests using status on deployment

ingeknudsen:


jonaspetersorensen:
You have a story for this so it will get done?
https://github.com/equinor/radix-operator/blob/f7a800fcd66595af6f9b5de77674e41d1ca7a104/pkg/apis/utils/deployment_builder.go#L84

jonaspetersorensen:
Is it ok to have radix version hardcoded?
https://github.com/equinor/radix-operator/blob/f7a800fcd66595af6f9b5de77674e41d1ca7a104/pkg/apis/utils/deployment_builder.go#L203

PR #299, opened 2019-09-02
Ra 753 secrets bug

ingeknudsen:


pkg/apis/utils/slice/sliceutils_test.go
```diff
@@ -0,0 +1,16 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Contains(t *testing.T) {
+	a := []string{"a", "b", "c"}
+
+	assert.True(t, ContainsString(a, "a"))
+	assert.True(t, ContainsString(a, "b"))
+	assert.True(t, ContainsString(a, "c"))
+	assert.False(t, ContainsString(a, "d"))
```
nemzes: Should add some tests to compare Unicode-normalised strings, e.g. the string with the bytes `61 30a` (`a +  ̊`) should be equal to the string with the single byte `e5` (`å`). Since this is user-provided input (secret names) we need to be cautious with Unicode.

PR #307, opened 2019-09-26
Remove restored status annotation

ingeknudsen:


nemzes:
Might want to add a comment about why this is needed (even just the Jira issue); it's not clear from the code.

PR #310, opened 2019-09-27
Ra 786 skip building when no build components

ingeknudsen:


pkg/apis/kube/namespaces.go
```diff
@@ -69,3 +71,40 @@ func (kube *Kube) ApplyNamespace(name string, annotations map[string]string, lab
...
+
+// NewNamespaceWatcherImpl Constructor
+func NewNamespaceWatcherImpl(client kubernetes.Interface) NamespaceWatcherImpl {
+	return NamespaceWatcherImpl{
+		client,
+	}
+}
+
+// WaitFor Waits for namespace to appear
+func (watcher NamespaceWatcherImpl) WaitFor(namespace string) error {
```
thezultimate: Error is never returned (except nil)

pkg/apis/kube/namespaces.go
```diff
@@ -69,3 +71,40 @@ func (kube *Kube) ApplyNamespace(name string, annotations map[string]string, lab
...
+	log.Infof("Waiting for namespace %s", namespace)
+	waitForNamespace(watcher.client, namespace)
+	log.Infof("Namespace %s exists and is active", namespace)
+	return nil
+
+}
+
+func waitForNamespace(client kubernetes.Interface, namespace string) {
+	for {
+		ns, _ := client.CoreV1().Namespaces().Get(namespace, metav1.GetOptions{})
```
thezultimate: Error can occur here. Maybe return error?

pkg/apis/kube/namespaces.go
ingeknudsen: Return error now, also on timeouts

pkg/apis/kube/namespaces.go
ingeknudsen: As above

PR #317, opened 2019-10-07
Ra 770 rolling restart

ingeknudsen:


radix-operator/deployment/controller.go
```diff
@@ -59,10 +61,28 @@ func NewController(client kubernetes.Interface,
...
+				logger.Debugf("Skip deployment object %s as it is inactive", newRD.GetName())
+				return
+			}
+
+			if deepEqual(oldRD, newRD) {
+				logger.Debugf("Deployment object is equal to old for %s. Do nothing", newRD.GetName())
+				return
+			}
+
 			controller.Enqueue(new)
```
nemzes: Maybe use an identifier other than `new`? At least GitHub thinks it's a reserved word (clearly isn't), but might make code reviews less confusing… 😛

radix-operator/deployment/controller.go
```diff
@@ -126,6 +146,16 @@ func NewController(client kubernetes.Interface,
 	return controller
 }
 
+func deepEqual(old, new *v1.RadixDeployment) bool {
+	if !reflect.DeepEqual(new.Spec, old.Spec) ||
+		!reflect.DeepEqual(new.ObjectMeta.Labels, old.ObjectMeta.Labels) ||
+		!reflect.DeepEqual(new.ObjectMeta.Annotations, old.ObjectMeta.Annotations) {
```
nemzes: `new`

radix-operator/job/controller.go
```diff
@@ -115,6 +136,16 @@ func NewController(client kubernetes.Interface,
 	return controller
 }
 
+func deepEqual(old, new *v1.RadixJob) bool {
```
nemzes: Can this function be generalised to avoid duplication? Maybe something like this (untested; no idea if syntax is correct):
```
func deepEqualK8sObject(old, new interface{}, t reflect.Type) bool {
	if !reflect.DeepEqual((*t)new.Spec, (*t)old.Spec) ||
		!reflect.DeepEqual((*t)new.ObjectMeta.Labels, (*t)old.ObjectMeta.Labels) ||
		!reflect.DeepEqual((*t)new.ObjectMeta.Annotations, (*t)old.ObjectMeta.Annotations) {
		return false
	}

	return true
}
```

radix-operator/deployment/controller.go
ingeknudsen: Renamed to cur

PR #327, opened 2019-10-16
Ra 814 set limit on rd and rj

ingeknudsen:


pkg/apis/deployment/deployment.go
```diff
@@ -423,3 +429,51 @@ func getLabelSelectorForComponent(component v1.RadixDeployComponent) string {
...
+		}
+
+		if len(allRDs.Items) > limit {
+			deployments := allRDs.Items
+			numToDelete := len(deployments) - limit
+			if numToDelete <= 0 {
+				return
+			}
+
+			sort.Sort(byActiveFrom(deployments))
```
keaaa: there already exist a sortRDsByActiveFromTimestampDesc that can be used for sorting

pkg/apis/job/job.go
```diff
@@ -481,3 +484,60 @@ func saveStatus(radixClient radixclient.Interface, rj *v1.RadixJob) error {
...
+		}
+
+		if len(allRJs.Items) > limit {
+			jobs := allRJs.Items
+			numToDelete := len(jobs) - limit
+			if numToDelete <= 0 {
+				return
+			}
+
+			sort.Sort(byCreated(jobs))
```
keaaa: again, there exist a sortJobsByActiveFromTimestampAsc that i think can be used

pkg/apis/deployment/deployment.go
ingeknudsen: Done

pkg/apis/job/job.go
ingeknudsen: Done

PR #341, opened 2019-10-29
Added service connection to US docker registry. Renamed service conne…

jonaspetersorensen:
…ction variables to better reflect what they represent

### DIFF
--- a/azure-pipelines/build-only-pipeline.yml
+++ b/azure-pipelines/build-only-pipeline.yml
@@ -10,8 +10,9 @@ pool:
   vmImage: 'Ubuntu-16.04'
 
 variables:
-  serviceConnectionDevSubscription: 'radix-azure-devops-dev-cr'
-  serviceConnectionProdSubscription: 'radix-azure-devops-prod-cr'
+  serviceConnectionDockerregistryDev: 'radix-azure-devops-dev-cr'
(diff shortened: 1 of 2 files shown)


azure-pipelines/build-only-pipeline.yml
```diff
@@ -10,8 +10,9 @@ pool:
   vmImage: 'Ubuntu-16.04'
 
 variables:
-  serviceConnectionDevSubscription: 'radix-azure-devops-dev-cr'
-  serviceConnectionProdSubscription: 'radix-azure-devops-prod-cr'
+  serviceConnectionDockerregistryDev: 'radix-azure-devops-dev-cr'
+  serviceConnectionDockerregistryProd: 'radix-azure-devops-prod-cr'
+  serviceConnectionDockerregistryUs: 'radix-azure-devops-us-cr'
```
thezultimate: Are these 3 lines used?

jonaspetersorensen: Not really, from what I can see. I added an extra reference to the new US repo, and renamed all the repo vars to be consistent with the other pipeline definition.
That said, now that I look closer at this "build only" pipeline I think Inge might want to refactor it a bit, because it seems to be hard coded to only use 1 specific service connection which I am not allowed to look at...

PR #343, opened 2019-10-29
fix-restart single component cascades to multiple

keaaa:


pkg/apis/deployment/environmentVariables_test.go
```diff
@@ -0,0 +1,28 @@
+package deployment
```
thezultimate: Make the file name consistent: `environmentvariables_test.go`

PR #366, opened 2019-12-13
Ra 510 deploy only

ingeknudsen:


pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -133,6 +133,33 @@ func PublicImageComponentCannotHaveSourceOrDockerfileSet(componentName string) e
...
+// ComponentWithDynamicTagRequiresTagInEnvironmentConfigForEnvironment Error if image is set with dynamic tag and tag is missing
+func ComponentWithDynamicTagRequiresTagInEnvironmentConfigForEnvironment(componentName, environment string) error {
+	return fmt.Errorf(
+		"Component %s with %s on image requires an image tag set on environment config for environment %s",
+		componentName, radixv1.DynamicTagNameInEnvironmentConfig, environment)
+}
+
+// ComponentWithDynamicTagCannotHaveTagInEnvironmentConfigForUnmappedEnvironment Unmapped environment cannot have tag
+func ComponentWithDynamicTagCannotHaveTagInEnvironmentConfigForUnmappedEnvironment(componentName, environment string) error {
+	return fmt.Errorf(
```
keaaa: is this an requirement - why does a branch need to be mapped to env? (e.g. deploy only)

pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -279,6 +306,27 @@ func validateComponents(app *radixv1.RadixApplication) []error {
 			errs = append(errs, PublicImageComponentCannotHaveSourceOrDockerfileSet(component.Name))
 		}
 
+		if component.Image != "" &&
```
keaaa: move into own method(s) - easier to understand whats being validated

pkg/apis/radixvalidators/validate_ra.go
ingeknudsen: Done

pkg/apis/radixvalidators/validate_ra.go
ingeknudsen: Done

PR #367, opened 2019-12-16
Ra 1128 add reconciled status to rr

ingeknudsen:


radix-operator/registration/controller_test.go
```diff
@@ -76,6 +76,10 @@ func Test_Controller_Calls_Handler(t *testing.T) {
 	assert.True(t, ok)
 	assert.True(t, op)
 
+	syncedRr, _ := radixClient.RadixV1().RadixRegistrations().Get(registration.GetName(), metav1.GetOptions{})
```
keaaa: maybe we should have a unit test that its not set when registration fails and before reconciling?

pkg/apis/application/application.go
```diff
@@ -99,7 +101,13 @@ func (app Application) OnSync() error {
 		return err
 	}
 
-	logger.Debugf("Applied access to ci/cd logs")
+	logger.Debugf("Applied access to ci/cd logs. Set registration to be reconciled")
+	radixRegistration.Status.Reconciled = metav1.NewTime(time.Now().UTC())
+	_, err = app.radixclient.RadixV1().RadixRegistrations().UpdateStatus(radixRegistration)
```
keaaa: Question: does updating status lead to a new reconciling (Update event)? Is this an issue?

ingeknudsen: Added deepEqual to controller code

PR #387, opened 2020-01-09
RA-1139 Deploy-only pipeline

thezultimate:


pipeline-runner/steps/deploy.go
```diff
@@ -44,63 +43,69 @@ func (cli *DeployStepImplementation) ErrorMsg(err error) string {
...
 // Deploy Handles deploy step of the pipeline
-func (cli *DeployStepImplementation) deploy(pipelineInfo *model.PipelineInfo) ([]v1.RadixDeployment, error) {
+func (cli *DeployStepImplementation) deploy(pipelineInfo *model.PipelineInfo) error {
 	appName := cli.GetAppName()
 	log.Infof("Deploying app %s", appName)
-	radixDeployments := []v1.RadixDeployment{}
 
-	for _, env := range pipelineInfo.RadixApplication.Spec.Environments {
-		if !deployment.DeployToEnvironment(env, pipelineInfo.TargetEnvironments) {
+	if !pipelineInfo.BranchIsMapped {
```
ingeknudsen: Should we consider changing this to a variable that makes sense in both circumstances? i.e. DeployConditionMet or probably a better naming in the same style???

thezultimate: I agree. This variable name becomes ambiguous. Changing it now.

PR #411, opened 2020-03-17
Added validation

ingeknudsen:


pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -478,6 +490,21 @@ func validateEnvironmentVariableNames(app *radixv1.RadixApplication) error {
...
+	for _, component := range app.Spec.Components {
+		for _, envConfig := range component.EnvironmentConfig {
+			for environmentVariable := range envConfig.Variables {
+				for _, secret := range component.Secrets {
+					if strings.EqualFold(environmentVariable, secret) {
+						return SecretNameConfictsWithEnvironmentVariable(component.Name, secret)
+					}
+				}
+			}
+		}
```
thezultimate: `for _, secret := range component.Secrets` are being called many times, which is not good performance wise if the `ra` contains a lot of secrets.

I would suggest for every component iteration, do a 1-pass iteration through `component.Secrets`, and save them to a map. Then `for environmentVariable := range envConfig.Variables` checks if it is available in that map or not.

ingeknudsen: See my updated code. Is that ok?

PR #419, opened 2020-03-26
Corrected patch

ingeknudsen:
Old:

image

After code change:
image

### DIFF
--- a/pkg/apis/applicationconfig/applicationconfig.go
+++ b/pkg/apis/applicationconfig/applicationconfig.go
@@ -16,7 +16,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/client-go/kubernetes"
 )
@@ -253,7 +252,8 @@ func (app *ApplicationConfig) applyEnvironment(newRe *v1.RadixEnvironment) error
 
 	repository := app.radixclient.RadixV1().RadixEnvironments()
 
-	oldRe, err := repository.Get(newRe.Name, metav1.GetOptions{})
+	// Get environment from cache, instead than for cluster
+	oldRe, err := app.kubeutil.GetEnvironment(newRe.Name)
 	if err != nil && errors.IsNotFound(err) {
 		// Environment does not exist yet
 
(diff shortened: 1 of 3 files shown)


pkg/apis/applicationconfig/applicationconfig.go
```diff
@@ -280,28 +280,38 @@ func (app *ApplicationConfig) applyEnvironment(newRe *v1.RadixEnvironment) error
 
 // patchDifference creates a mergepatch, comparing old and new RadixEnvironments and issues the patch to radix
 func patchDifference(repository radixTypes.RadixEnvironmentInterface, oldRe *v1.RadixEnvironment, newRe *v1.RadixEnvironment, logger *log.Entry) error {
+	radixEnvironment := oldRe.DeepCopy()
```
JoakimHagen: leave a comment on why this object is copied and not mutated

pkg/apis/applicationconfig/applicationconfig.go
```diff
@@ -280,28 +280,38 @@ func (app *ApplicationConfig) applyEnvironment(newRe *v1.RadixEnvironment) error
...
 	}
 
-	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldReJSON, newReJSON, v1.RadixEnvironment{})
+	patchBytes, err := strategicpatch.CreateTwoWayMergePatch(oldReJSON, radixEnvironmentJSON, v1.RadixEnvironment{})
 
-	if string(patchBytes) != "{}" {
-		patchedLimitRange, err := repository.Patch(newRe.Name, types.StrategicMergePatchType, patchBytes)
+	if !isEmptyPatch(patchBytes) {
+		// Will perform update as patching does not seem to work for this custom resource
+		patchedEnvironment, err := repository.Update(radixEnvironment)
```
JoakimHagen: As I read the doc for Update when I wrote my code, I understood it as updating the radixEnvironment object by writing the current cluster resource data into it. Not the other way around, which Patch does. Am I wrong?

ingeknudsen: Patch gives an error. I don't think StrategicMergePatchType works for CRDs. I am not sure how you made it work, but an existing namespace were never patched

pkg/apis/applicationconfig/applicationconfig.go
ingeknudsen: This pattern was taken from an example of writing a controller. At this time I cannot find that. This is the pattern we have opted for in all other cases. I would rather ask you, why you think deepcopy should be avoided, backed up with some references why we should drop deepCopy and do mutation instead. Is it memory leaks that you are afraid of? If you give a good argument for not using deepCopy, we can maybe pull this into a separate US.

JoakimHagen: I find adding several lines of code without knowing why we do so is going to hurt us more than help. It might be the correct thing to do, but when I don't understand why it's there I'm bound to make mistakes elsewhere, adding it where it just creates noise or leaving it out where it should be.

ingeknudsen: See this article:

https://insujang.github.io/2020-02-13/programming-kubernetes-crd/

Look at the comment:
  // NEVER modify objects directly (it is in read-only local cache.)
  // We use DeepCopy() and modify this copy.

This is especially true now that we retrieve object from cache, and not from cluster (in your original code). Retrieving from cache speeds up the operator. Now it takes much less time to reconcile a cluster. If everything is as it should be, then reconciling should be lightning fast. Before we started retrieving objects from cache, it took the operator nearly 20 minutes to run through reconciliation of the playground cluster, before it was properly up to date.

PR #427, opened 2020-04-03
RA-1265 orphaned re status

JoakimHagen:


radix-operator/environment/controller.go
```diff
@@ -145,6 +146,33 @@ func NewController(client kubernetes.Interface,
...
+
+			for _, envName := range droppedEnvironments(oldRa, newRa) {
+				uniqueName := utils.GetEnvironmentNamespace(oldRa.Name, envName)
+				re, err := radixClient.RadixV1().RadixEnvironments().Get(uniqueName, metav1.GetOptions{})
+				if err != nil {
+					controller.Enqueue(re)
+				}
+			}
+		},
+		DeleteFunc: func(cur interface{}) {
```
ingeknudsen, line 165: Maybe this is unnecessary, as users will never be able to delete an RA, so it is no use case.

JoakimHagen: Perhaps. I think I'd rather keep it for the mental decoupling of expected user permissions. I don't want to come back to it and be confused about apparent holes in its behavior.

PR #430, opened 2020-04-21
rbac tables added

JoakimHagen:


docs/RBAC.md
```diff
@@ -4,6 +4,8 @@
 
 ![pic](diagrams/platform_user.png)
 
+See table 1.3 for complete listing of permissions
```
thezultimate, line 7: This figure (`platform_user.png`) is missing role and rolebinding `*-machine-user-token` inside the `*-app` namespace that gives `*-machine-user service account` access to secret containing access token.

thezultimate, line 7: Why use `clusterrolebinding` for `*-machine-user` service account instead of `rolebinding`? The service account exists inside `*-app` namespace. A link to the service account is also missing in the picture.

Consistency would be good here since `rolebinding` `radix-app-admin` is used (not `clusterrolebinding`) to give `*-machine-user` access to `clusterrole` `radix-app-admin`.

JoakimHagen: "Why use clusterrolebinding for *-machine-user service account instead of rolebinding?"
I do not know. I document it as it is. We can fix and optimize now that we see these issues.

JoakimHagen: What tool is used to make these images? Does it have a non-rasterized source?

JoakimHagen: @thezultimate What about the radix-private-image-hubs, radix-app-admin-build-secrets and <app>-machine-user-token roles that aren't mentioned in the diagram? should they all be added?

thezultimate: Maybe create a user story card for this improvement? To make it consistent.

docs/RBAC.md
```diff
@@ -224,3 +179,118 @@
...
+radix-pipeline-runner|radix|radixapplications|radixapplications|radixapplications|radixapplications|radixapplications||
+radix-platform-user|k8s|selfsubjectaccessreviews|namespaces, jobs, ingresses, horizontalpodautoscalers|namespaces, jobs, ingresses, horizontalpodautoscalers|namespaces, jobs, ingresses, horizontalpodautoscalers|||
+radix-platform-user|radix|radixregistrations|radixjobs|radixapplications, radixdeployments, radixjobs|radixapplications, radixjobs|radixjobs||
+radix-platform-user-rr-`application`|radix||radixregistrations|radixregistrations|radixregistrations|radixregistrations|radixregistrations|radixregistrations
+radix-private-image-hubs|k8s||secrets|secrets|secrets|secrets|secrets|secrets
+radix-webhook|k8s|jobs|namespaces, ingresses, deployments, jobs|namespaces, ingresses, deployments, jobs|namespaces, ingresses, deployments, jobs||jobs|
+radix-webhook|radix|radixjobs|radixregistrations, radixapplications, radixdeployments, radixjobs|radixregistrations, radixapplications, radixdeployments, radixjobs|radixregistrations, radixapplications, radixdeployments, radixjobs||radixjobs|
+
+Role|Domain|All permissions
```
thezultimate, line 293: This part doesn't render properly. Typo?

PR #439, opened 2020-05-04
Common component resource settings

thezultimate:


pipeline-runner/steps/promotion.go
```diff
@@ -185,6 +186,11 @@ func mergeWithRadixApplication(radixConfig *v1.RadixApplication, radixDeployment
 				radixDeployment.Spec.Components[index].EnvironmentVariables[variableKey] = variableValue
 			}
 		}
+
+		// Append common resources settings if currently empty
+		if reflect.DeepEqual(radixDeployment.Spec.Components[index].Resources, v1.ResourceRequirements{}) {
```
keaaa, line 191: For promotion will Component.Resource take precedence over Component.Environemt[x].Resource? shouldn't it be other way around?

pipeline-runner/steps/promotion_test.go
```diff
@@ -133,6 +133,13 @@ func TestPromote_PromoteToOtherEnvironment_NewStateIsExpected(t *testing.T) {
 							WithName("app").
 							WithCommonEnvironmentVariable("DB_TYPE", "mysql").
 							WithCommonEnvironmentVariable("DB_NAME", "my-db").
+							WithCommonResource(map[string]string{
+								"memory": "64Mi",
+								"cpu":    "250m",
+							}, map[string]string{
+								"memory": "128Mi",
+								"cpu":    "500m",
+							}).
```
keaaa, line 142: maybe also test the precedence

pkg/apis/deployment/radixcomponent.go
```diff
@@ -45,6 +46,11 @@ func getRadixComponentsForEnv(radixApplication *v1.RadixApplication, containerRe
 			}
 		}
 
+		// Append common resources settings if currently empty
+		if reflect.DeepEqual(resources, v1.ResourceRequirements{}) {
```
keaaa, line 50: same here - environmentspecificconfig should take precedence over common component settings

pkg/apis/deployment/radixcomponent_test.go
```diff
@@ -227,6 +227,50 @@ func TestGetRadixComponentsForEnv_CommonEnvironmentVariables_With_Override(t *te
 	assert.Equal(t, "environment_common_2_dev_override", deployComponentDev[1].EnvironmentVariables["ENV_COMMON_2"])
 }
 
+func TestGetRadixComponentsForEnv_CommonResources(t *testing.T) {
```
keaaa, line 230: test precedence

thezultimate: Precedence is tested in line 260-263.

pipeline-runner/steps/promotion.go
thezultimate: No, Component.Environemt[x].Resource takes precedence over Component.Resource.

The logic is:
(1) Set value to Component.Environemt[x].Resource
(2) If (1) is empty, set value to Component.Resource

We use the standard DeepEqual function to check if a struct is empty.

pipeline-runner/steps/promotion_test.go
thezultimate: I will update the test.

PR #447, opened 2020-05-26
RA-1416 Fix Env Orphan flag on creating Application

JoakimHagen:


charts/radix-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: radix-operator
-version: 1.0.39
-appVersion: 1.5.11
+version: 1.0.41
+appVersion: 1.5.13
```
thezultimate, line 4: Maybe `1.5.12`?

charts/radix-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: radix-operator
-version: 1.0.39
-appVersion: 1.5.11
+version: 1.0.41
```
thezultimate, line 3: Maybe `1.0.40`?

PR #448, opened 2020-05-29
adds metric for resource requests

keaaa:


keaaa:
uncertain about the move of logic into struct def - radixdeploytypes.go - any comments?

PR #457, opened 2020-06-25
Enforce lowercase name in radixconfig

olemoi:
Applications with uppercase /metadata/name in radixconfig.yaml will trigger a warning in the build process stating that the name contains uppercase letters and automatically convert it to lowercase for  the user.

### DIFF
--- a/pkg/apis/radixvalidators/validate_ra_test.go
+++ b/pkg/apis/radixvalidators/validate_ra_test.go
@@ -50,7 +50,6 @@ func Test_application_name_casing_is_validated(t *testing.T) {
 		{"Upper case name", radixvalidators.ApplicationNameNotLowercaseError(upperCaseName), func(ra *v1.RadixApplication) { ra.Name = upperCaseName }},
 	}
 
-	// _, client := validRASetup()
 	for _, testcase := range testScenarios {
 		t.Run(testcase.name, func(t *testing.T) {
 			validRA := createValidRA()
--- a/pkg/apis/radixvalidators/validate_ra_test.go
+++ b/pkg/apis/radixvalidators/validate_ra_test.go
@@ -1,6 +1,7 @@
 package radixvalidators_test
 
 import (
+	"strings"
 	"testing"
 
 	v1 "github.com/equinor/radix-operator/pkg/apis/radix/v1"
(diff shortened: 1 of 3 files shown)


pkg/apis/radixvalidators/validate_ra_test.go
```diff
@@ -32,6 +33,43 @@ func Test_missing_rr(t *testing.T) {
...
+
+			if err != nil {
+				assert.False(t, isValid)
+				assert.NotNil(t, err)
+				assert.True(t, testcase.expectedError.Error() == err.Error())
+				assert.True(t, strings.ToLower(validRA.Name) == expectedName)
+			} else {
+				assert.True(t, isValid)
+				assert.Nil(t, err)
+			}
```
JoakimHagen, line 68: maybe also add `assert.True(t, validRA.Name == expectedName)`?

pkg/apis/radixvalidators/validate_ra_test.go
```diff
@@ -32,6 +33,43 @@ func Test_missing_rr(t *testing.T) {
...
+		name          string
+		expectedError error
+		updateRa      updateRAFunc
+	}{
+		{"Mixed case name", radixvalidators.ApplicationNameNotLowercaseError(mixedCaseName), func(ra *v1.RadixApplication) { ra.Name = mixedCaseName }},
+		{"Lower case name", radixvalidators.ApplicationNameNotLowercaseError(lowerCaseName), func(ra *v1.RadixApplication) { ra.Name = lowerCaseName }},
+		{"Upper case name", radixvalidators.ApplicationNameNotLowercaseError(upperCaseName), func(ra *v1.RadixApplication) { ra.Name = upperCaseName }},
+	}
+
+	// _, client := validRASetup()
```
JoakimHagen, line 53: is this old code?

pkg/apis/radixvalidators/validate_ra_test.go
olemoi: The validator does not actively convert the name, it just checks if the name contains uppercase characters. The conversion takes place in `apply_radixconfig.go`

pkg/apis/radixvalidators/validate_ra_test.go
olemoi: Yes, removed.

pkg/apis/radixvalidators/validate_ra_test.go
JoakimHagen: You're right.

PR #464, opened 2020-08-05
Revert "Set target env only if present (#463)"

satr:
This reverts commit 41ba2a0f7099442e080ac604819d78168818d85b.
Revert tag also?


satr:
Obsolete

PR #467, opened 2020-08-18
Ra 1076 environment label

olemoi:
Supply RJ with target environments from RA if RA exists upon creation of RJ. Sync environments between RJ and RA every time RJ is updated.

### DIFF
--- a/radix-operator/job/controller.go
+++ b/radix-operator/job/controller.go
@@ -1,8 +1,6 @@
 package job
 
 import (
-	"reflect"
-
 	"github.com/equinor/radix-operator/pkg/apis/job"
 	"github.com/equinor/radix-operator/pkg/apis/kube"
 
@@ -74,6 +72,7 @@ func NewController(client kubernetes.Interface,
 	jobInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(cur interface{}) {
 			radixJob, _ := cur.(*v1.RadixJob)
+			job.AddTargetEnvironments(radixJob)
 			if job.IsRadixJobDone(radixJob) {
 				logger.Debugf("Skip job object %s as it is complete", radixJob.GetName())
 				metrics.CustomResourceAddedButSkipped(crType)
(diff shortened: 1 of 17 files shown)


radix-operator/job/controller.go
```diff
@@ -79,11 +81,40 @@ func NewController(client kubernetes.Interface,
 				return
 			}
 
+			radixApplication, err := radixClient.RadixV1().RadixApplications(radixJob.Namespace).Get(radixJob.Spec.AppName, metav1.GetOptions{})
```
thezultimate, line 84: Extract the new changes to a function (maybe in `job.go`?).

radix-operator/job/controller.go
```diff
@@ -79,11 +81,40 @@ func NewController(client kubernetes.Interface,
...
+			}
+			radixJob.Status.TargetEnvs = targetEnvs
+			radixClient.RadixV1().RadixJobs(radixJob.Namespace).UpdateStatus(radixJob)
+
 			controller.Enqueue(cur)
 			metrics.CustomResourceAdded(crType)
 		},
 		UpdateFunc: func(old, cur interface{}) {
 			newRJ := cur.(*v1.RadixJob)
+
```
thezultimate, line 102: Extract the new changes in a function (maybe `job.go`?).

radix-operator/job/controller.go
```diff
@@ -79,11 +81,40 @@ func NewController(client kubernetes.Interface,
 				return
 			}
 
+			radixApplication, err := radixClient.RadixV1().RadixApplications(radixJob.Namespace).Get(radixJob.Spec.AppName, metav1.GetOptions{})
+			var targetEnvs []string
+
+			if err != nil {
+				targetEnvs = append(targetEnvs, "N/A")
```
thezultimate, line 88: `err` can be many things (not only object not found). Maybe better handle this? You can use `if errors.IsNotFound`.

radix-operator/job/controller.go
```diff
@@ -79,11 +81,40 @@ func NewController(client kubernetes.Interface,
...
+			radixApplication, err := radixClient.RadixV1().RadixApplications(newRJ.Namespace).Get(newRJ.Spec.AppName, metav1.GetOptions{})
+
+			if err == nil {
+				// Radix application exists
+				var raEnvsSlice []string
+				for _, env := range radixApplication.Spec.Environments {
+					raEnvsSlice = append(raEnvsSlice, env.Name)
+				}
+
+				if !reflect.DeepEqual(newRJ.Status.TargetEnvs, raEnvsSlice) {
```
thezultimate, line 112: Is there a case where this block is not entered?

radix-operator/job/controller.go
olemoi: If there is an error fetching RA, TargetEnvironments can not be update anyway, figured best way to handle this is set env to N/A for all situations it can't be determined.

thezultimate (approved):

PR #480, opened 2020-09-04
RA-1768 added validation for non-empty WBS

satr:


pkg/apis/radixvalidators/validate_rr.go
```diff
@@ -26,6 +26,11 @@ func AppNameCannotBeEmptyError() error {
 	return ResourceNameCannotBeEmptyError("app name")
 }
 
+// WbsCannotBeEmptyError WBS cannot be empty
+func WbsCannotBeEmptyError() error {
```
keaaa, line 30: is this used anywhere?

satr: No, this is outdated method. I deleted it.
Validation is in `validate_rr.validateWbs()` now

keaaa (approved):

PR #507, opened 2020-12-09
Supported multiple blobfuse volumes for same storage account

satr:
https://dev.azure.com/Equinor/Radix/_workitems/edit/58726

### DIFF
--- a/pkg/apis/radixvalidators/validate_ra_test.go
+++ b/pkg/apis/radixvalidators/validate_ra_test.go
@@ -576,6 +576,7 @@ func Test_ValidationOfVolumeMounts_Errors(t *testing.T) {
 				volumeMounts := []v1.RadixVolumeMount{
 					{
 						Type:      "disk",
+						Name:      "some_name",
 						Container: "some_container_name",
 						Path:      "some_path",
 					},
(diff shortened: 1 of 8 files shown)


pkg/apis/radixvalidators/validate_ra_test.go
```diff
@@ -584,18 +585,79 @@ func Test_ValidationOfVolumeMounts_Errors(t *testing.T) {
 			},
 			false,
 			false,
+			"not recognized volume mount type",
+		},
+		{
+			"blob mount type with different containers",
+			func(ra *v1.RadixApplication) {
+				volumeMounts := []v1.RadixVolumeMount{
```
ingeknudsen, line 593: Should rather be:

```
    - name: component
      environmentConfig:
        - environment: dev
          volumeMounts:
            - type: blob
              containers:
                - container: some_container_name_1
                  path: some_path_1
                - container: some_container_name_2
                  path: some_path_2
```

ingeknudsen (changes requested):
Suggested change is to have a list of container below mount

ingeknudsen (dismissed):
Ah, now I see your final edits. I would not have list of containers, if this is not requested by Anders. If what Anders ask for is to mount a-storage-account, a-container to one path and  b-storage-account, c-container to another path, then that is what I think we should deliver in the first iteration. Then if some team ask for mapping multiple containers two the same storage account, Radix team can either decide to implement what you suggest above or just tell the team that they can get around that by (but then I guess Radix team needs to ensure there are two secrets per distinct name):

```
    - name: web
      environmentConfig:
        - environment: dev
          volumeMounts:
            - type: blob
              name: a-storage-account
              container: a-container
              path: /usr/share/nginx/html/blob1
            - type: blob
              name: a-storage-account
              container: another-container
              path: /usr/share/nginx/html/blob2
```

It is important no to implement full flexibility if it is not asked for. I will approve your PR, because it looks good to me, then you can decide :slightly_smiling_face:

ingeknudsen (approved):

PR #520, opened 2021-02-22
72391 extend ra and rd

nilsgstrabo:
- Extend RA and RD with jobs
- Handle RA jobs in radix-pipeline => jobs in RD


satr (approved):
Looks good

PR #523, opened 2021-03-02
Refactor garbage collector for deploymenthandler in radix-operator

nilsgstrabo:
- garbage collection use the radix-component label to verfiy that a kubernetes object is owned by radix.
- compare agains both components and jobs in RD during gargabe collection


satr (approved):
Looks good

PR #531, opened 2021-03-11
Run all deployed applications as non root

olemoi:


pkg/apis/deployment/radixcomponent.go
```diff
@@ -30,6 +30,8 @@ func getRadixComponentsForEnv(radixApplication *v1.RadixApplication, env string,
 		var imageTagName string
 
 		var alwaysPullImageOnDeploy bool
+		// Containers run as non-root unless overridden in config
+		runAsNonRoot := true
```
nilsgstrabo: Should be false

nilsgstrabo (changes requested):
-Add runAsNonRoot to jobComponents
-Add test for nonRoot when building RadixApplication => RadixDeployment

nilsgstrabo (approved):

PR #562, opened 2021-04-23
add certification to configuration

Pespiri:


pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -105,6 +106,10 @@ func CPUResourceRequirementFormatError(value string) error {
 	return fmt.Errorf("Format of cpu resource requirement %s (value %s) is wrong. Must match regex '%s'", "cpu", value, cpuRegex)
 }
 
+func InvalidValidationType(verification string) error {
```
nilsgstrabo, line 109: Name should be InvalidVerificationType

pkg/apis/radixvalidators/validate_ra_test.go
```diff
@@ -155,19 +155,19 @@ func Test_invalid_ra(t *testing.T) {
 		{"too long environment variable name", radixvalidators.InvalidStringValueMaxLengthError("environment variable name", wayTooLongName, 253), func(ra *v1.RadixApplication) {
 			ra.Spec.Components[1].EnvironmentConfig[0].Variables[wayTooLongName] = "Any value"
 		}},
-		{"conflicting variable and secret name", radixvalidators.SecretNameConflictsWithEnvironmentVariable(validRASecondComponentName, conflicingVariableName), func(ra *v1.RadixApplication) {
```
nilsgstrabo, line 158: I think we should add authentication config to the radixconfig.yaml used to test a valid RA (in line 97)
image

nilsgstrabo (changes requested):
Looks good! Just two minor changes that needs to be adressed

pkg/apis/radixvalidators/validate_ra.go
Pespiri: Fixed

pkg/apis/radixvalidators/validate_ra_test.go
Pespiri: Added

nilsgstrabo (approved):

PR #568, opened 2021-06-03
88595 csi volume driver

satr:


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/equinor/radix-operator
 go 1.13
 
 require (
+	github.com/pkg/errors v0.9.1 // indirect
```
nilsgstrabo, line 6: VS Code complains when `github.com/pkg/errors` is in the list. A `go mod tidy` removes it.

satr: Fixed

pkg/apis/deployment/jobschedulercomponent.go
```diff
@@ -65,7 +65,7 @@ func (js *jobSchedulerComponent) GetResources() *v1.ResourceRequirements {
 }
 
 func (js *jobSchedulerComponent) GetVolumeMounts() []v1.RadixVolumeMount {
```
nilsgstrabo, line 67: This method does the same as it's "base" struct (v1.RadixDeployJobComponent) and can therefore be removed from jobSchedulerComponent.

satr: Fixed

pkg/apis/deployment/volumemount.go
```diff
@@ -105,19 +220,441 @@ func (deploy *Deployment) createOrUpdateVolumeMountsSecrets(namespace, component
...
-	secrets, err := deploy.listSecretsForForBlobVolumeMount(component)
+	_, err := deploy.kubeutil.ApplySecret(namespace, &secret)
 	if err != nil {
 		return err
 	}
 
+	return nil
+}
+
+func (deploy *Deployment) garbageCollectSecrets(secrets []*v1.Secret, excludeSecretNames []string) error {
```
nilsgstrabo, line 252: Feels a bit strange to have a method with such a generic name in volumemount.go. And it is only used by the garbageCollectVolumeMountsSecretsNoLongerInSpecForComponent method.

pkg/apis/deployment/volumemount.go
```diff
@@ -105,19 +220,441 @@ func (deploy *Deployment) createOrUpdateVolumeMountsSecrets(namespace, component
...
+		err := deploy.DeletePersistentVolume(pv.Name)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+//CreateOrUpdateCsiAzureResources Create or update CSI Azure volume resources - StorageClasses, PersistentVolumeClaims, PersistentVolume
+func (deploy *Deployment) CreateOrUpdateCsiAzureResources(desiredDeployment *appsv1.Deployment) error {
```
nilsgstrabo, line 451: What was the reason you need to use the k8s Deployment object, and not a RadixCommonDeployComponent, to create/update a storgae class and pvc?

nilsgstrabo (dismissed):

nilsgstrabo (approved):

PR #571, opened 2021-06-08
102320 refactored GetPodAffinity, made BuildResourceRequirements public

Pespiri:


pkg/apis/utils/affinity.go
```diff
@@ -11,25 +11,25 @@ import (
...
 	return nil
 }
 
-func addGpuNodeSelectorTerms(deployComponent v1.RadixCommonDeployComponent, nodeAffinity *corev1.NodeAffinity) {
+func addGpuNodeSelectorTerms(node *v1.RadixNode, nodeAffinity *corev1.NodeAffinity) {
 	nodeSelectorTerm := corev1.NodeSelectorTerm{}
-	addNodeSelectorRequirementForGpu(deployComponent.GetNode().Gpu, &nodeSelectorTerm)
-	addNodeSelectorRequirementForGpuCount(deployComponent.GetNode().GpuCount, &nodeSelectorTerm)
+	addNodeSelectorRequirementForGpu(node.Gpu, &nodeSelectorTerm)
+	addNodeSelectorRequirementForGpuCount(node.GpuCount, &nodeSelectorTerm)
```
nilsgstrabo, line 32: Should add a nil check for `node`.

nilsgstrabo (changes requested):
Missing a nil check. Was missing before too, but best to add to be on the safe side.

pkg/apis/utils/affinity.go
Pespiri: fixed

PR #579, opened 2021-06-23
102114 get common env vars for radix components

satr:
For radix-operator - it is used config-map and DNS_ZONE env-var;
For radix-app (like radix-job-scheduler) - only env-vars used, and RADIX_DNS_ZONE
Class pkg/apis/utils/environmentvariables_set.go is to collect EnvVars, avoid repetition on adding, return list with same order, as EnvVar-s were added

### DIFF
--- a/pkg/apis/deployment/environmentvariables.go
+++ b/pkg/apis/deployment/environmentvariables.go
@@ -13,10 +13,55 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func GetEnvironmentVariablesFrom(appName string, kubeutil *kube.Kube, radixDeployment *v1.RadixDeployment, deployComponent v1.RadixCommonDeployComponent) []corev1.EnvVar {
+type envVariablesSourceDecorator interface {
+	getClusterName() (string, error)
+	GetContainerRegistry() (string, error)
+}
+
+type radixApplicationEnvVariablesSourceDecorator struct{}
+type radixOperatorEnvVariablesSourceDecorator struct {
+	kubeutil *kube.Kube
+}
+
+func (envVariablesSource *radixApplicationEnvVariablesSourceDecorator) getClusterName() (string, error) {
+	return os.Getenv(defaults.ClusternameEnvironmentVariable), nil
+}
+
+func (envVariablesSource *radixApplicationEnvVariablesSourceDecorator) GetContainerRegistry() (string, error) {
+	return os.Getenv(defaults.ContainerRegistryEnvironmentVariable), nil
+}
+
+func (envVariablesSource *radixOperatorEnvVariablesSourceDecorator) getClusterName() (string, error) {
+	clusterName, err := envVariablesSource.kubeutil.GetClusterName()
+	if err != nil {
+		return "", fmt.Errorf("failed to get cluster name from ConfigMap: %v", err)
+	}
+	return clusterName, nil
+}
+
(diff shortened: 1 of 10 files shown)


pkg/apis/deployment/environmentvariables.go
```diff
@@ -86,51 +140,34 @@ func appendAppEnvVariables(radixDeployName string, radixEnvVars v1.EnvVarsMap) [
...
+		return envVarSet.Items()
 	}
 
-	clusterType := os.Getenv(defaults.OperatorClusterTypeEnvironmentVariable)
-	if clusterType != "" {
-		environmentVariables = append(environmentVariables, corev1.EnvVar{
-			Name:  defaults.RadixClusterTypeEnvironmentVariable,
-			Value: clusterType,
-		})
+	clusterType, err := getEnvVar(defaults.RadixClusterTypeEnvironmentVariable)
```
nilsgstrabo, line 151: The cluster type value must also be added to the environmentVariablesSourceDecorator interface.
Operator reads the value from RADIXOPERATOR_CLUSTER_TYPE variable, and job scheduler must read it from RadixClusterTypeEnvironmentVariable.
The unit tests in deployment_test does not set all radix-operator variables, like RADIXOPERATOR_CLUSTER_TYPE, which I think is a bug

pkg/apis/utils/environmentvariables_set.go
```diff
@@ -0,0 +1,64 @@
...
+//NewEnvironmentVariablesSet Create new EnvironmentVariablesSet
+func NewEnvironmentVariablesSet() *environmentVariablesSet {
+	return (&environmentVariablesSet{}).Init(make([]corev1.EnvVar, 0))
+}
+
+//Init Init set with environment variables list
+func (set *environmentVariablesSet) Init(environmentVariables []corev1.EnvVar) *environmentVariablesSet {
+	(*set).items = environmentVariables
+	(*set).newItems = make([]*corev1.EnvVar, 0)
+	(*set).itemsMap = getMap(&environmentVariables)
```
nilsgstrabo, line 22: Don't think you need to wrap `set `in `(*set)`.

pkg/apis/utils/environmentvariables_set.go
```diff
@@ -0,0 +1,64 @@
...
+
+//Get Get an environment variable by name
+func (set *environmentVariablesSet) Get(name string) (corev1.EnvVar, bool) {
+	if envVar, ok := set.itemsMap[name]; ok {
+		return *envVar, true
+	}
+	return corev1.EnvVar{}, false
+}
+
+func getMap(environmentVariables *[]corev1.EnvVar) map[string]*corev1.EnvVar {
```
nilsgstrabo, line 57: environmentVariables doesn't have to be a pointer to slice as long as you are not modifying it (append, delete).

nilsgstrabo (changes requested):
Need to update the deployment unit test for common variables to include all the RADIX_* variables

pkg/apis/utils/environmentvariables_set.go
satr: Fixed

pkg/apis/utils/environmentvariables_set.go
satr: Fixed

pkg/apis/deployment/environmentvariables.go
satr: Added logic

nilsgstrabo (dismissed):

PR #594, opened 2021-08-24
Fix deployment with error during cm creation

satr:
Fixed config-map sync - removed lister from config-map getter. 
Added error handling for create desired deployment

### DIFF
--- a/pkg/apis/kube/config_maps.go
+++ b/pkg/apis/kube/config_maps.go
@@ -5,18 +5,21 @@ import (
 	"encoding/json"
 	"fmt"
 	log "github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/strategicpatch"
-
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
 // CreateConfigMap Create config map
 func (kubeutil *Kube) CreateConfigMap(namespace string, configMap *corev1.ConfigMap) (*corev1.ConfigMap, error) {
-	return kubeutil.kubeClient.CoreV1().ConfigMaps(namespace).Create(context.TODO(),
+	createdConfigMap, err := kubeutil.kubeClient.CoreV1().ConfigMaps(namespace).Create(context.TODO(),
 		configMap,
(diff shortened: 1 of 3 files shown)


pkg/apis/kube/config_maps.go
```diff
@@ -5,37 +5,33 @@ import (
...
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 )
 
 // CreateConfigMap Create config map
 func (kubeutil *Kube) CreateConfigMap(namespace string, configMap *corev1.ConfigMap) (*corev1.ConfigMap, error) {
-	return kubeutil.kubeClient.CoreV1().ConfigMaps(namespace).Create(context.TODO(),
+	createdConfigMap, err := kubeutil.kubeClient.CoreV1().ConfigMaps(namespace).Create(context.TODO(),
```
nilsgstrabo, line 16: You can just return directly. Are there situations where Get returns an error and a configmap, or vice versa; nil error and nil configmap?

pkg/apis/kube/config_maps.go
```diff
@@ -5,37 +5,33 @@ import (
...
-		configMap, err = kubeutil.kubeClient.CoreV1().ConfigMaps(namespace).Get(context.TODO(), name, metav1.GetOptions{})
-		if err != nil {
-			return nil, err
-		}
-		log.Debugf("Created environment variables ConfigMap  '%s'", configMap.GetName())
+	configMap, err = kubeutil.kubeClient.CoreV1().ConfigMaps(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
 	}
+	log.Debugf("Created environment variables ConfigMap  '%s'", configMap.GetName())
```
nilsgstrabo, line 34: Debug message (Create) does not match operation (Get).
Guess you can return directly from the call to Get. Ref comment above

nilsgstrabo (changes requested):
A few comments related to Get and Create configmap methods

nilsgstrabo (approved):
A few minor comments, but nothing important

pkg/apis/kube/config_maps.go
satr: Fixed

pkg/apis/kube/config_maps.go
satr: Removed

PR #603, opened 2021-09-03
109742 extend unit tests for editable env vars

satr:


pkg/apis/test/utils.go
```diff
@@ -36,6 +36,10 @@ func NewTestUtils(client kubernetes.Interface, radixclient radixclient.Interface
 	}
 }
 
+func (tu *Utils) GetKubeUtil() *kube.Kube {
```
nilsgstrabo, line 39: Not used in radix-operator. Is it used in other projects?

nilsgstrabo (approved):

PR #604, opened 2021-09-07
Added access for user to config-maps via api

satr:


charts/radix-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: radix-operator
 version: 1.1.5
```
nilsgstrabo, line 3: Perhaps we need to change version (chart)?

nilsgstrabo (changes requested):
Perhaps it is version, and not appVersion that must be changed?

nilsgstrabo (approved):

charts/radix-operator/Chart.yaml
satr: Fixed

PR #620, opened 2021-11-08
Job steps order

oleterno-equinor:
image


nilsgstrabo (commented):
Would it be better to do the sorting in the Radix Web Console?
Each step has a timestamp for when it started. If two steps has the same started timestamp, we can do a secondary sort by name to prevent items from "jumping". If a step does not have a value for started (don't know if this will happen, and if it happens it should only be for a short time), we can put it at the end.
Sorting it in front-end will also be backward compatible for old jobs.

PR #638, opened 2022-01-11
104262 csi driver for azure keyvault

satr:


charts/radix-operator/templates/deployment.yaml
```diff
@@ -99,6 +99,8 @@ spec:
               value: {{ .Values.logLevel }}
             - name: RADIX_CONTAINER_REGISTRY
               value: {{ .Values.containerRegistry }}
+            - name: RADIXOPERATOR_TENANT_ID
```
nilsgstrabo, line 102: Should the name be more specific to its use, e.g. RADIXOPERATOR_SECRETPROVIDER_TENANT_ID?

charts/radix-operator/templates/radix-apps-rbac.yaml
```diff
@@ -51,6 +51,13 @@ rules:
   - watch
   - patch
   - create
+- apiGroups:
+    - 'secrets-store.csi.x-k8s.io'
+  resources:
+    - secretproviderclasses
+  verbs:
+    - get
+    - list
```
nilsgstrabo, line 60: Why does radix-webhook need access to secretproviderclasses?

charts/radix-operator/templates/radix-pipeline-rbac.yaml
```diff
@@ -42,6 +42,17 @@ rules:
...
+    - 'secrets-store.csi.x-k8s.io'
+  resources:
+    - secretproviderclasses
+  verbs:
+    - get
+    - list
+    - create
+    - update
+    - patch
+    - delete
```
nilsgstrabo, line 55: Why does radix-pipeline need access to secretproviderclasses?

operator.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.16.9-alpine3.14 as builder
+FROM golang:alpine3.15 as builder
```
nilsgstrabo, line 1: Should we use a tag that is specific to the go version we are using, like `golang:1.17-alpine3.15`?

pkg/apis/utils/init.go
```diff
@@ -24,7 +24,7 @@ var (
 )
 
 // GetKubernetesClient Gets clients to talk to the API
-func GetKubernetesClient() (kubernetes.Interface, radixclient.Interface, monitoring.Interface) {
+func GetKubernetesClient() (kubernetes.Interface, radixclient.Interface, monitoring.Interface, *secretProviderClient.Clientset) {
```
nilsgstrabo, line 27: The other return types are of type` something.Interface`, should a `secretProviderClient.Interface` be returned insteade of a `*secretProviderClient.Clientset`?

charts/radix-operator/templates/deployment.yaml
satr: I think as there is no explicit reason so far to have separate tenant-id for the secret provider, we can use Radix tenant-id. If we'll face such need for separate tenant-id for the secret provider - we can create separate variable and use it.

charts/radix-operator/templates/radix-apps-rbac.yaml
satr: Mistake - it supposed to be in radix-api - (moved now). I'm not 100% sure it is required in radix-api before testing as a service.

charts/radix-operator/templates/radix-pipeline-rbac.yaml
satr: Removed all but get/list

operator.Dockerfile
satr: Corrected to `golang:1.17.6-alpine3.15`

pkg/apis/test/utils.go
```diff
@@ -28,7 +29,7 @@ type Utils struct {
 
 // NewTestUtils Constructor
 func NewTestUtils(client kubernetes.Interface, radixclient radixclient.Interface) Utils {
-	kubeUtil, _ := kube.New(client, radixclient)
+	kubeUtil, _ := kube.New(client, radixclient, secretproviderfake.NewSimpleClientset())
```
nilsgstrabo, line 32: The secretproviderfake should be a parameter to NewTestUtils.

pipeline-runner/steps/promotion.go
```diff
@@ -187,6 +187,7 @@ func mergeJobComponentsWithRadixApplication(radixConfig *v1.RadixApplication, ra
 		}
 
 		newEnvJob.Secrets = job.Secrets
+		newEnvJob.SecretRefs = job.SecretRefs
```
nilsgstrabo, line 190: Why is SecretRefs copied from the source RD component? Isn't SecretRefs rebuilt based on the environment we promote to?

pipeline-runner/steps/promotion.go
```diff
@@ -209,6 +210,7 @@ func mergeComponentsWithRadixApplication(radixConfig *v1.RadixApplication, radix
 		}
 
 		newEnvComponent.Secrets = component.Secrets
+		newEnvComponent.SecretRefs = component.SecretRefs
```
nilsgstrabo, line 213: Why is SecretRefs copied from the source RD component? Isn't SecretRefs rebuilt based on the environment we promote to?

pipeline-runner/steps/promotion.go
nilsgstrabo: Perhaps the unit test should include a component where secretRefs is set in environmentConfig?

pipeline-runner/steps/promotion.go
satr: Agree. Removed. I think `newEnvJob.Secrets = job.Secrets` is also wrong

pipeline-runner/steps/promotion.go
satr: Agree. Removed. I think newEnvJob.Secrets = job.Secrets is also wrong

satr: > Perhaps the unit test should include a component where secretRefs is set in environmentConfig?

I'll create those unit-tests

pkg/apis/utils/init.go
satr: Correct. Fixed in all cases

charts/radix-operator/templates/deployment.yaml
nilsgstrabo: Agree. Perhaps I can use it in oauth proxy to build the default token issuer URL.

pkg/apis/deployment/deployment.go
```diff
@@ -46,16 +46,11 @@ type Deployment struct {
 	registration            *v1.RadixRegistration
 	radixDeployment         *v1.RadixDeployment
 	securityContextBuilder  SecurityContextBuilder
+	tenantId                string
```
nilsgstrabo, line 49: I see that other methods that use the environment variables in the operator reads them directly with os.GetEnv, or it is wrapped in methods in api/defauls that also reads directly by using os.GetEnv.
You inject the value for tenantId in the "constructor", and I like this better than reading directly from env.
I think we should use a common pattern for getting values from env vars, perhaps define a struct for the various values from env vars where tenantId will be one of the fields. 
We don't have to do it now tough. I think it is fine for now to inject tenantId, and we can do the rest later.

satr: I agree - this connection to env-vars is really bad practice, problems with testing, even worse then startic global vars. We have Env object in other projects, maybe we can use similar (with framework you utilized some time ago).

pkg/apis/deployment/environmentvariables.go
```diff
@@ -98,10 +98,10 @@ func getEnvironmentVariablesFrom(kubeutil *kube.Kube, appName string, envVarsSou
...
+	return getEnvironmentVariables(kubeutil, appName, envVarsSource, radixDeployment, deployComponent.GetName(), deployComponent.GetSecrets(), deployComponent.GetSecretRefs(), isPortPublic, deployComponent.GetPorts(), envVarsConfigMap, deployComponent.GetEnvironmentVariables())
 }
 
-func getEnvironmentVariables(appName string, envVarsSource environmentVariablesSourceDecorator, radixDeployment *v1.RadixDeployment, componentName string, radixSecretNames []string, isPublic bool, ports []v1.ComponentPort, envVarConfigMap *corev1.ConfigMap, deployComponentEnvVars v1.EnvVarsMap) []corev1.EnvVar {
+func getEnvironmentVariables(kubeutil *kube.Kube, appName string, envVarsSource environmentVariablesSourceDecorator, radixDeployment *v1.RadixDeployment, componentName string, radixSecretNames []string, radixSecretRefs v1.RadixSecretRefs, isPublic bool, ports []v1.ComponentPort, envVarConfigMap *corev1.ConfigMap, deployComponentEnvVars v1.EnvVarsMap) ([]corev1.EnvVar, error) {
```
nilsgstrabo, line 104: kubeutil is not used and should be removed

nilsgstrabo, line 104: the error return values is always nil. Should be removed.

satr: Fixed

satr: Fixed

pkg/apis/deployment/radixcommoncomponent.go
```diff
@@ -22,3 +26,130 @@ func updateComponentNode(component v1.RadixCommonComponent, node *v1.RadixNode)
...
+		if len(commonAzureKeyVault.Items) == 0 { //use all from environments for this Azure Key Vault
+			continue
+		}
+		envItemEnvVarsMap := make(map[string]v1.RadixAzureKeyVaultItem)
+		envAzureKeyVaultItems := envAzureKeyVault.Items
+		for _, item := range envAzureKeyVaultItems {
+			envItemEnvVarsMap[item.EnvVar] = item
+		}
+		for _, keyVaultItem := range commonAzureKeyVault.Items {
+			if _, existsInEnvSecretRefs := envItemEnvVarsMap[keyVaultItem.EnvVar]; !existsInEnvSecretRefs {
```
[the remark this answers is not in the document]
satr: if-block on which line?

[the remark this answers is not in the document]
satr: That's right, this check is not needed. Refined a bit code.

pkg/apis/deployment/volumemount.go
```diff
@@ -109,18 +161,85 @@ func getRadixVolumeTypeIdForName(radixVolumeMountType radixv1.MountType) string
...
+					return nil, fmt.Errorf("missing Azure Key vault name in the secret provider class %s", secretProviderClass.Name)
+				}
+				credsSecretName := defaults.GetCsiAzureKeyVaultCredsSecretName(deployComponent.GetName(), azKeyVaultName)
+				volume.VolumeSource.CSI = &corev1.CSIVolumeSource{
+					Driver:               csiSecretStoreDriver,
+					ReadOnly:             commonUtils.BoolPtr(true),
+					VolumeAttributes:     map[string]string{csiVolumeSourceVolumeAttrSecretProviderClassName: secretProviderClass.Name},
+					NodePublishSecretRef: &corev1.LocalObjectReference{Name: credsSecretName},
+				}
+				break
```
nilsgstrabo, line 224: break is not needed

pkg/apis/deployment/volumemount.go
```diff
@@ -109,18 +161,85 @@ func getRadixVolumeTypeIdForName(radixVolumeMountType radixv1.MountType) string
...
+	for _, azureKeyVault := range secretRef.AzureKeyVaults {
+		secretProviderClassName := kube.GetComponentSecretProviderClassName(deployComponent.GetName(), radixDeploymentName, radixv1.RadixSecretRefTypeAzureKeyVault, azureKeyVault.Name)
+		secretProviderClass, err := kubeutil.GetSecretProviderClass(namespace, secretProviderClassName)
+		if err != nil {
+			return nil, err
+		}
+		for _, secretObject := range secretProviderClass.Spec.SecretObjects {
+			volume := v1.Volume{
+				Name:         secretObject.SecretName,
+				VolumeSource: v1.VolumeSource{},
```
nilsgstrabo, line 208: VolumeSource is not nilable, so you don't need to initialise it with an empty map

pkg/apis/kube/secret_provider.go
```diff
@@ -0,0 +1,83 @@
...
+	radixv1 "github.com/equinor/radix-operator/pkg/apis/radix/v1"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	secretsstorev1 "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
+	"strings"
+)
+
+// StringArray ...
+type StringArray struct {
```
nilsgstrabo, line 16: This type is only used in deployment/secrets.go. Should it be moved there instead and made private by naming it stringArray?

pkg/apis/deployment/secrets.go
```diff
@@ -138,6 +138,182 @@ func (deploy *Deployment) createOrUpdateSecretsForComponent(registration *radixv
...
+			Encoding: commonUtils.StringUnPtr(item.Encoding),
+		}
+		if item.Type != nil {
+			obj.Type = string(*item.Type)
+		} else {
+			obj.Type = string(radixv1.RadixAzureKeyVaultObjectTypeSecret)
+		}
+		parameterObject = append(parameterObject, obj)
+	}
+	parameterObjectArray := kube.StringArray{Array: []string{}}
```
nilsgstrabo, line 244: No need to initialise an array. kube.StringArrar{} is the same.

pkg/apis/kube/secret_provider.go
```diff
@@ -0,0 +1,83 @@
...
+	// Format. Optional. The format of the Azure Key Vault object, supported types are pem and pfx. objectFormat: pfx is only supported with objectType: secret and PKCS12 or ECC certificates. Default format for certificates is pem.
+	Format string `yaml:"objectFormat,omitempty"`
+	// Encoding. Optional. Setting object encoding to base64 and object format to pfx will fetch and write the base64 decoded pfx binary
+	Encoding string `yaml:"objectEncoding,omitempty"`
+}
+
+// GetSecretProviderClass Gets secret provider class
+func (kubeutil *Kube) GetSecretProviderClass(namespace string, className string) (*secretsstorev1.SecretProviderClass, error) {
+	class, err := kubeutil.secretProviderClient.SecretsstoreV1().SecretProviderClasses(namespace).Get(context.Background(), className, metav1.GetOptions{})
+	if err != nil {
```
nilsgstrabo, line 39: I think it is better to always return an error, even if it is IsNotFound. It is better for the caller to handle decide what to do with IsNotFound than to return a nil value for SecretProviderClass and error when it is IsNotFound.

pkg/apis/kube/secret_provider.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+	return class, err
+}
+
+// ListSecretProviderClass Gets secret provider classes for the component
+func (kubeutil *Kube) ListSecretProviderClass(namespace string, componentName string) ([]secretsstorev1.SecretProviderClass, error) {
+	classList, err := kubeutil.secretProviderClient.SecretsstoreV1().SecretProviderClasses(namespace).
+		List(context.Background(), metav1.ListOptions{LabelSelector: GetAllSecretRefObjectsLabelSelector(componentName)})
+	if err != nil {
```
nilsgstrabo, line 52: Same as for GetSecretProviderClass. Return the error and let the caller decide.

pkg/apis/radix/v1/radixdeploytypes.go
```diff
@@ -290,10 +316,12 @@ func (deployComponent RadixDeployComponent) GetNrOfReplicas() int32 {
 // The job component is used by the radix-job-scheduler-server to create Kubernetes Job objects
 type RadixDeployJobComponent struct {
 	Name                    string                    `json:"name" yaml:"name"`
+	Environment             string                    `json:"environment" yaml:"environment"`
```
nilsgstrabo, line 319: Environment seems to be unused. Why do we need it? The RD has a label with the environment name.

pkg/apis/deployment/volumemount.go
satr: Fixed

pkg/apis/kube/secret_provider.go
satr: Moved

pkg/apis/deployment/secrets.go
satr: Fixed

pkg/apis/kube/secret_provider.go
satr: So far we did not returned `errors.IsNotFound` error, but something like this
```
if errors.IsNotFound(err) {
            return nil, nil
```
This approach (with `errors.IsNotFound`) seems to me clear, and we can start using it, if you suggested this.

pkg/apis/radix/v1/radixdeploytypes.go
satr: Removed. I think it was used at some point in the `GetRadixComponentsForEnv` methods
No it is used `deployComponent.SecretRefs = getRadixCommonComponentRadixSecretRefs(&radixComponent, environmentSpecificConfig)`

pkg/apis/radix/v1/radixdeploytypes.go
```diff
@@ -312,9 +340,11 @@ type RadixCommonDeployComponent interface {
 	GetPorts() []ComponentPort
 	GetEnvironmentVariables() EnvVarsMap
 	GetSecrets() []string
+	GetSecretRefs() RadixSecretRefs
 	GetMonitoring() bool
 	GetResources() *ResourceRequirements
 	GetVolumeMounts() []RadixVolumeMount
+	GetEnvironment() string
```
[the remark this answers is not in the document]
satr: Removed

pkg/apis/radix/v1/radixdeploytypes.go
```diff
@@ -328,6 +358,7 @@ type RadixCommonDeployComponent interface {
 	GetAuthentication() *Authentication
 	SetName(name string)
 	SetVolumeMounts(mounts []RadixVolumeMount)
+	SetEnvironment(environment string)
```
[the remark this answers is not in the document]
satr: Removed

pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -700,100 +721,225 @@ func validateSecrets(app *radixv1.RadixApplication) error {
...
+			}
+			if err := validateVariableName("Azure Key vault secret references name", keyVaultItem.Name); err != nil {
 				return err
 			}
 		}
 	}
+	return nil
+}
+
+func validateSecretRefsPath(component v1.RadixCommonComponent) error {
```
nilsgstrabo, line 837: This funtion is never used

pkg/apis/kube/secret_provider.go
nilsgstrabo: All other Get/List methods for Kube returns any error without checking the type of error. I just feel that they should behave the same way.

pkg/apis/radix/v1/radixdeploytypes.go
```diff
@@ -223,6 +232,10 @@ func (deployJobComponent *RadixDeployJobComponent) GetVolumeMounts() []RadixVolu
 	return deployJobComponent.VolumeMounts
 }
 
+func (deployJobComponent *RadixDeployJobComponent) GetEnvironment() string {
```
nilsgstrabo, line 235: This should be removed too?

nilsgstrabo (approved):

pkg/apis/kube/secret_provider.go
satr: Fixed

pkg/apis/kube/secret_provider.go
satr: Fixed

pkg/apis/radixvalidators/validate_ra.go
satr: Removed

pkg/apis/radix/v1/radixdeploytypes.go
satr: Removed

PR #639, opened 2022-01-12
added timeLimitSeconds to radixconfig. passed as ActiveDeadlineSeconds

magnus-longva-bouvet:


pkg/apis/deployment/kubedeployment.go
```diff
@@ -182,7 +182,6 @@ func (deploy *Deployment) setDesiredDeploymentProperties(deployComponent v1.Radi
 	desiredDeployment.Spec.Template.Spec.Containers[0].Image = deployComponent.GetImage()
 	desiredDeployment.Spec.Template.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
 	desiredDeployment.Spec.Template.Spec.Containers[0].SecurityContext = deploy.securityContextBuilder.BuildContainerSecurityContext(deployComponent)
-	desiredDeployment.Spec.Template.Spec.Containers[0].ImagePullPolicy = corev1.PullAlways
```
magnus-longva-bouvet, line 185: Removed this line because it appears to be duplicated on 183

pkg/apis/deployment/kubedeployment.go
```diff
@@ -226,6 +225,11 @@ func (deploy *Deployment) updateDeploymentByComponent(deployComponent v1.RadixCo
 		desiredDeployment.Spec.Replicas = int32Ptr(int32(DefaultReplicas))
 	}
 
+	timeLimitSeconds := deployComponent.GetTimeLimitSeconds()
```
magnus-longva-bouvet, line 228: If I've understood correctly, this function is the appropriate place to add logic which only applies to either regular components or job components. True?

satr (changes requested):

satr (approved):

PR #650, opened 2022-02-07
185626 move scan to last step in pipeline

Pespiri:


nilsgstrabo (changes requested):
Update appVersion version in chart.yaml

nilsgstrabo (approved):

PR #651, opened 2022-02-09
use api version v1 for CustomResourceDefinitions

oleterno-equinor:


nilsgstrabo (changes requested):
Update app and chart version in chart.yaml

nilsgstrabo (approved):

PR #665, opened 2022-02-22
changed egressPort.Number to egressPort.Port

magnus-longva-bouvet:


pkg/apis/networkpolicy/networkpolicy.go
```diff
@@ -61,7 +61,7 @@ func convertToK8sEgressRule(radixEgressRule rx.EgressRule) (egressRule v1.Networ
 		ports = append(ports, v1.NetworkPolicyPort{
 			Protocol: &portProtocol,
 			Port: &intstr.IntOrString{
-				IntVal: radixPort.Number,
+				IntVal: radixPort.Port,
```
satr, line 64: `validate_ra` still have method name `validateEgressRulePortNumber(number int32)`

satr (changes requested):

pkg/apis/networkpolicy/networkpolicy.go
magnus-longva-bouvet: I think the name still makes sense. What should the method be renamed to?

satr: Just suggestion to rename to `validateEgressRulePort(port **int32)`

PR #667, opened 2022-02-24
Set unique name for the radix-clone-config job

satr:


nilsgstrabo (changes requested):
I think we also should append the hash to the build and scan jobs. These steps are not used by the deploy only job type, but if someone manages to create two build-deploy jobs on the same second, we could end up with the same conflict (I think)

nilsgstrabo (approved):

PR #673, opened 2022-03-02
115432 extend monitoring config in radixconfig

Pespiri:


pkg/apis/deployment/servicemonitor.go
```diff
@@ -14,8 +14,13 @@ import (
 )
 
 func (deploy *Deployment) createOrUpdateServiceMonitor(deployComponent v1.RadixCommonDeployComponent) error {
+	monitoringConfig := deployComponent.GetMonitoringConfig()
+	if monitoringConfig.PortName == "" {
+		monitoringConfig.PortName = deployComponent.GetPorts()[0].Name
```
satr, line 19: I would recommend to check if the ports array length is > 0. As you can see here (avoid issue with sidecar-components in feature)
```
func getReadinessProbeForComponent(component v1.RadixCommonDeployComponent) (*corev1.Probe, error) {
	if len(component.GetPorts()) == 0 {
		return nil, nil
	}
```

satr (changes requested):

pkg/apis/deployment/servicemonitor.go
Pespiri: done

satr (approved):

PR #682, opened 2022-04-22
added RADIX_ZONE to default pod env vars

magnus-longva-bouvet:


pkg/apis/deployment/environmentvariables.go
```diff
@@ -80,6 +86,10 @@ func (envVarsSource *radixOperatorEnvironmentVariablesSourceDecorator) getCluste
 	return egressIps, nil
 }
 
+func (envVarsSource *radixOperatorEnvironmentVariablesSourceDecorator) getRadixZone() (string, error) {
+	return getEnvVar(defaults.RadixZoneEnvironmentVariable)
+}
+
```
magnus-longva-bouvet, line 92: I have used exact same implementation of getRadixZone for both radixOperatorEnvironmentVariablesSourceDecorator and radixApplicationEnvironmentVariablesSourceDecorator. Problem?

satr: I do not see problem here

pkg/apis/defaults/environment_variables.go
```diff
@@ -92,4 +92,7 @@ const (
 
 	// KubernetesApiPortEnvironmentVariable Port which the K8s API server listens to for HTTPS
 	KubernetesApiPortEnvironmentVariable = "KUBERNETES_SERVICE_PORT"
+
+	// RadixZoneEnvironmentVariable denotes the Radix zone of the current cluster. E.g. "c1" or "c2"
+	RadixZoneEnvironmentVariable = "RADIX_ZONE"
```
satr, line 97: Collision with `RADIX_ZONE` with another purpose. What is a purpose of it?

satr (changes requested):

PR #690, opened 2022-04-27
replace random with least_conn

magnus-longva-bouvet:


magnus-longva-bouvet:
`The Random load balancing method should be used for distributed environments where multiple load balancers are passing requests to the same set of backends. For environments where the load balancer has a full view of all requests, use other load balancing methods, such as round robin, least connections and least time.`

https://docs.nginx.com/nginx/admin-guide/load-balancer/http-load-balancer/#choosing-a-load-balancing-method

Considering that we do have multiple nginx backends, it looks like the `random` load balancing method most closely resembles the intuitive interpretation of "leastconnectedlb".

PR #703, opened 2022-06-03
209965 controller parallell processing

magnus-longva-bouvet:


nilsgstrabo (changes requested):
changes as discussed

nilsgstrabo (approved):
Approved.

nilsgstrabo (changes requested):
Use environment variables (in deployment) for the chart values. Easier to override when debugging locally.
Should we use default values for #threads be 1, and QPS+burts 5?

PR #723, opened 2022-07-15
adding git hash and git tags as RD annotations

magnus-longva-bouvet:


pkg/apis/kube/kube.go
```diff
@@ -18,6 +18,8 @@ import (
 // Radix Annotations
 const (
 	RadixBranchAnnotation          = "radix-branch"
+	RadixGitTagsAnnotation         = "radix-git-tags"
+	RadixCommitAnnotation          = "radix-commit"
```
satr, line 22: I suggest we start using this naming convention in annotations, we agreed recently. It matches better to k8s [naming convention](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set), even though "The prefix is optional", it is usually in use.
`PipelineNameAnnotation = "radix.equinor.com/tekton-pipeline-name"`
https://github.com/equinor/radix-tekton/blob/ea684d6553c894a230ca3ba770cb0a4c77b05f78/pkg/defaults/pipelines.go#L5

satr (changes requested):

satr (dismissed):

anneliawa (approved):

PR #759, opened 2022-09-08
Make PDB when HPA.MinReplicas > 1

anneliawa:


magnus-longva-bouvet (approved):

pkg/apis/deployment/pdb.go
```diff
@@ -21,15 +20,27 @@ func getNumberOfReplicas(component v1.RadixCommonDeployComponent) int {
 
 }
 
-func (deploy *Deployment) createOrUpdatePodDisruptionBudget(component v1.RadixCommonDeployComponent) error {
-	namespace := deploy.radixDeployment.Namespace
-	componentName := component.GetName()
-
+func componentShallHavePdb(component v1.RadixCommonDeployComponent) bool {
```
nilsgstrabo, line 23: There is a special condition described in the docs (https://www.radix.equinor.com/references/reference-radix-config/#environmentconfig), when replicas=0 then a HPA is not created.
This condition must also be considered when creating the PDB. When replicas=0, no PDB must be created

nilsgstrabo (changes requested):
Added a few comments that must be adressed

nilsgstrabo (approved):

PR #768, opened 2022-10-06
230155 add cache env var

magnus-longva-bouvet:


nilsgstrabo (approved):

charts/radix-operator/values.yaml
```diff
@@ -18,6 +18,7 @@ configToMap: radix-config-2-map:master-latest
 imageBuilder: radix-image-builder:master-latest
 jobScheduler: radix-job-scheduler-server:main-latest
 radixTekton: radix-tekton:main-latest
+useImageBuilderCache: xx
```
nilsgstrabo, line 21: We should have a valid default

nilsgstrabo (changes requested):

nilsgstrabo (dismissed):

nilsgstrabo (approved):

PR #769, opened 2022-10-07
Allow radixconfig in subfolder and custom name

satr:


pkg/apis/defaults/pipelinejobs.go
```diff
@@ -1,6 +1,16 @@
...
+// RadixPipelineJobPipelineContainerName The container name of the Radix pipeline orchestration job
 const RadixPipelineJobPipelineContainerName = "radix-pipeline"
+
+// RadixPipelineJobPreparePipelinesContainerName The container name of the pipeline job, reading the RadixApplication from the Radix configuration file and preparing the Sub-pipleine, if it is configured
 const RadixPipelineJobPreparePipelinesContainerName = "prepare-pipelines"
+
+// RadixPipelineJobRunPipelinesContainerName The container name of the pipeline job, running the Sub-pipleine, if it is configured
 const RadixPipelineJobRunPipelinesContainerName = "run-pipelines"
+
+// RadixCacheLayerNamePrefix The name of the cache container image
```
magnus-longva-bouvet, line 12: It's technically not an image
```suggestion
// RadixCacheLayerNamePrefix The name of the cache artifact
```

pkg/apis/radixvalidators/validate_rr.go
```diff
@@ -209,3 +214,15 @@ func validateConfigBranch(name string) error {
 
 	return nil
 }
+
+// ValidateRadixConfigFullName Validates the radixconfig file name and path
+func ValidateRadixConfigFullName(radixConfigFullName string) error {
```
magnus-longva-bouvet, line 219: I understand this method will only be used in radix-api. Do unit tests also belong in radix-api?

satr: It was initially only in radix-api and moved here, because it gets used in api-s different packages. I thought to put tests here, which would duplicate API tests. Maybe it is ok though

magnus-longva-bouvet (approved):

PR #772, opened 2022-10-10
release PR

magnus-longva-bouvet:


satr (changes requested):
I think first we need to add this change and test it
https://equinor.slack.com/archives/G9M0R6BSB/p1665152704125929

satr (approved):

PR #774, opened 2022-10-21
Removed run-as-non-root

anneliawa:


pkg/apis/deployment/securitycontext.go
```diff
@@ -2,40 +2,33 @@ package deployment
...
 }
 
 type securityContextBuilder struct {
 	forceRunAsNonRoot bool
 }
 
-func NewSecurityContextBuilder(forceRunAsNonRoot bool) *securityContextBuilder {
-	return &securityContextBuilder{forceRunAsNonRoot: forceRunAsNonRoot}
+func NewSecurityContextBuilder() *securityContextBuilder {
+	return &securityContextBuilder{forceRunAsNonRoot: true}
```
satr, line 18: Argument should not be removed

satr (changes requested):

pkg/apis/deployment/securitycontext.go
anneliawa: should this still not be removed?

satr: As it is decided to remove functionality `forceRunAsNonRoot` - this need to be removed (you did this allredy correctly), as well as arguments in logic it used (and env-var)

satr (changes requested):
This can be removed https://github.com/equinor/radix-operator/blob/7ae5c7dbb367b98ca404eba21bde1c2069ab4604/pkg/apis/defaults/environment_variables.go#L91.
Also any use of `RADIX_DEPLOYMENTS_FORCE_NON_ROOT_CONTAINER`

satr (approved):

PR #784, opened 2022-11-16
Add to tekton permits to get namespace

satr:


pkg/apis/kube/namespaces.go
```diff
@@ -128,7 +128,7 @@ func waitForNamespace(client kubernetes.Interface, namespace string) error {
 	for {
 		go func() {
 			ns, err := client.CoreV1().Namespaces().Get(context.TODO(), namespace, metav1.GetOptions{})
-			if !k8errs.IsNotFound(err) {
+			if !k8errs.IsNotFound(err) && !k8errs.IsForbidden(err) {
```
nilsgstrabo, line 131: Could we refactor this to use  wait.PollImmediateUntilWithContext?

nilsgstrabo (dismissed):
Apporved, but perhaps we should consider refactoring the waitForNamespace() function (see my comment)

nilsgstrabo (approved):

pkg/apis/kube/namespaces.go
satr: fixed

PR #848, opened 2023-03-17
199772 Move SSH private key from RR to Secret

anneliawa:


radix-operator/registration/controller.go
```diff
@@ -126,9 +126,29 @@ func NewController(client kubernetes.Interface,
 		},
 	})
 
+	configMapInformer := kubeInformerFactory.Core().V1().ConfigMaps()
+	configMapInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			cm, converted := obj.(*corev1.ConfigMap)
+			if !converted {
+				logger.Errorf("corev1.ConfigMap object cast failed during deleted event received.")
+				// TODO: Configmap was deleted, but not recreated.
```
magnus-longva-bouvet, line 135: This scenario has happened at least once during testing. Why does this some times happen, and how can we recover from it? The consequence is that a user clicks the "Regenerate deploy key button", but a new key does _not_ appear in web-console. In order to get a new deploy key, the user would need to make some arbitrary change to RR and trigger a resync

pkg/apis/utils/keyGen.go
magnus-longva-bouvet, line 1: This file is pretty much copy-pasted and moved from radix-api

radix-operator/registration/controller.go
nilsgstrabo: I don't know the details, but for some reason a deleted object is either the object itself, or a cache.DeletedFinalStateUnknown. This struct type has an .Obj property which is the actual deleted object.
You need to try casting to both types.

You can see how it is used here: [https://github.com/equinor/radix-operator/blob/2e77bd637ac19167fd5b3b56f8c4605f28aaa079/radix-operator/common/controller.go#L222](https://github.com/equinor/radix-operator/blob/2e77bd637ac19167fd5b3b56f8c4605f28aaa079/radix-operator/common/controller.go#L222)

pkg/apis/utils/keyGen.go
nilsgstrabo: Should consider switching to ed22519 ("crypto/ed25519") instead of rsa?

pkg/apis/defaults/configmaps.go
```diff
@@ -9,4 +9,8 @@ const (
 	RadixGitChangedFolders = "git-changed-folders"
 	// RadixGitChangedChangedRadixConfig Indicates if Radix config file is changed in the GitHub source
 	RadixGitChangedChangedRadixConfig = "git-changed-radixconfig"
+	GitPublicKeyConfigMapName         = "git-ssh-public-key"
+	GitPublicKeyConfigMapKey          = "public-key"
+	GitPrivateKeySecretName           = "git-ssh-keys"
+	GitPrivateKeySecretKey            = "id-rsa"
```
nilsgstrabo, line 15:
```suggestion
	GitPrivateKeySecretKey            = "id_rsa"
```

The existing key is "id_rsa" (underscore). GitPrivateKeySecretKey must use the same key name to prevent regeneration of all keys.

pkg/apis/defaults/configmaps.go
```diff
@@ -9,4 +9,8 @@ const (
 	RadixGitChangedFolders = "git-changed-folders"
 	// RadixGitChangedChangedRadixConfig Indicates if Radix config file is changed in the GitHub source
 	RadixGitChangedChangedRadixConfig = "git-changed-radixconfig"
+	GitPublicKeyConfigMapName         = "git-ssh-public-key"
+	GitPublicKeyConfigMapKey          = "public-key"
```
nilsgstrabo, line 13:
```suggestion
	GitPublicKeyConfigMapKey          = "public_key"
```
Use underscore to align with name of private key

nilsgstrabo (changes requested):
Added some suggestions and one issue that must be fixed

nilsgstrabo (approved):

PR #851, opened 2023-03-17
Add pipeline job option image tags

satr:


nilsgstrabo (changes requested):
Remove the unused Env mock. Otherwise approved.

PR #926, opened 2023-09-08
Removed antiaffinity from jobs, reduced builder resources

satr:


pkg/apis/batch/kubejob.go
```diff
@@ -63,9 +62,9 @@ func (s *syncer) validatePayloadSecretReference(batchJob *radixv1.RadixBatchJob,
 	if err != nil {
 		return err
 	}
-	if !radixlabels.GetRadixBatchDescendantsSelector(jobComponent.GetName()).Matches(labels.Set(payloadSecret.GetLabels())) {
-		return fmt.Errorf("secret %s, referenced in the job %s of the batch %s is not valid payload secret", batchJob.PayloadSecretRef.Name, batchJob.Name, s.batch.GetName())
-	}
+	// if !radixlabels.GetRadixBatchDescendantsSelector(jobComponent.GetName()).Matches(labels.Set(payloadSecret.GetLabels())) {
```
nilsgstrabo, line 65: Why is this code commented out?

nilsgstrabo (changes requested):

pkg/apis/batch/kubejob.go
satr: It should be uncommented. I’ll fix it

satr: Fixed

nilsgstrabo (approved):

PR #947, opened 2023-10-09
add batch and pipeline job pod annotation to prevent eviction by cluster autoscaler

nilsgstrabo:
Add annotation "cluster-autoscaler.kubernetes.io/safe-to-evict: false" to batch and pipeline job pods.
This annotation prevents the cluster autoscaler from removing the Pod's node when determining if a node can be scaled down.

Ref: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-the-parameters-to-ca

### DIFF
--- a/charts/radix-operator/Chart.yaml
+++ b/charts/radix-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.23.4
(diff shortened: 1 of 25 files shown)


satr (approved):

Richard87 (approved):
lgtm!

PR #948, opened 2023-10-10
Change logic on seccomp copy task

sveinpj:


charts/radix-operator/templates/seccomp-profile-daemonset.yaml
```diff
@@ -24,7 +24,7 @@ spec:
         - name: file-copy-container
           image: alpine
           command: ["sh", "-c"]
-          args: ["while true; do cp /configmap-data/{{ .Values.seccompProfile.fileNameOnNode }} /hostpath/; sleep 20; done"]
+          args: ["while true; do if ! diff -q /configmap-data/{{ .Values.seccompProfile.fileNameOnNode }} /hostpath/{{ .Values.seccompProfile.fileNameOnNode }} > /dev/null 2>&1; then echo "$(date ''+%H:%M'') - Changes" && cp /configmap-data/{{ .Values.seccompProfile.fileNameOnNode }} /hostpath/; else echo "$(date ''+%H:%M'') - No changes"; fi; sleep 60; done"]
```
nilsgstrabo, line 27:
```suggestion
          args: ["while true; do if ! diff -q /configmap-data/{{ .Values.seccompProfile.fileNameOnNode }} /hostpath/{{ .Values.seccompProfile.fileNameOnNode }} > /dev/null 2>&1; then echo \"$(date ''+%H:%M'') - Changes\" && cp /configmap-data/{{ .Values.seccompProfile.fileNameOnNode }} /hostpath/; else echo \"$(date ''+%H:%M'') - No changes\"; fi; sleep 60; done"]
```
Escape quotes

nilsgstrabo (changes requested):
Fix missing quote escape

nilsgstrabo (approved):

PR #960, opened 2023-10-19
use AddAfter for requeue when lock cannot be acquired

nilsgstrabo:
Using AddAfter when requeuing an item when a lock cannot be acquired. AddRateLimited could potentially lead to situations when an item would be delayed for 1000 seconds when requeuing.


Richard87 (approved):
lgtm

PR #986, opened 2023-11-24
Update validation of imageTagName in CICD pipeline

nilsgstrabo:
- Do not fail deploy step in pipeline job when component/job image to deploy does not contain {imageTagName} and radixconfig has imageTagName set. This will happen when promoting.
- Fail deploy pipeline if destination environment contains components to build
- Validate components/jobs in buildeploy or deploy pipeline with {imageTagName} has imageTagName set in radixconfig or supplied as pipline argument

### DIFF
--- a/charts/radix-operator/Chart.yaml
+++ b/charts/radix-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.25.4
-appVersion: 1.45.4
+version: 1.25.5
+appVersion: 1.45.5
 kubeVersion: ">=1.24.0"
 description: Radix Operator
 keywords:
--- a/charts/radix-operator/Chart.yaml
+++ b/charts/radix-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.25.6
-appVersion: 1.45.6
+version: 1.25.7
+appVersion: 1.45.7
 kubeVersion: ">=1.24.0"
 description: Radix Operator
 keywords:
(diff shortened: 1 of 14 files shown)


charts/radix-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.25.4
-appVersion: 1.45.4
+version: 1.25.5
+appVersion: 1.45.5
```
Richard87, line 4: Må bumpes til 7 eller over...

pipeline-runner/steps/errors.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/pkg/errors"
+)
+
+var (
+	ErrDeployOnlyPipelineDoesNotSupportBuild = errors.New("deploy pipeline does not support building components and jobs")
+)
+
+func ErrMissingRequiredImageTagName(componentName, envName string) error {
+	return fmt.Errorf("component %s requires imageTagName to be set in environmentConfig for environment %s or specified as argument to the pipeline job", componentName, envName)
```
Richard87, line 14: Kan deler av feilmeldingen være statisk slik at Error.Is fungerer?

Richard87 (changes requested):
noen kommentarer

pipeline-runner/steps/errors.go
nilsgstrabo: fixed

charts/radix-operator/Chart.yaml
nilsgstrabo: fixed

Richard87 (approved):
godkjent

PR #996, opened 2023-11-30
Fix promote error when imagetagname set

nilsgstrabo:


emirgens:
Can branch be deleted? "...branch has unmerged commits"

nilsgstrabo:
Yes, delete it

PR #1028, opened 2024-01-02
Remove use of lister from apply roles

satr:


pkg/apis/applicationconfig/environment.go
```diff
@@ -0,0 +1,101 @@
...
+		WithAppName(app.config.Name).
+		WithAppLabel().
+		WithEnvironmentName(env.Name).
+		WithRegistrationOwner(app.registration).
+		WithEgressConfig(env.Egress).
+		// Orphaned flag will be set by the environment handler but until
+		// reconciliation we must ensure it is false
+		// Update: It seems Update method does not update status object when using real k8s client, but the fake client does.
+		// Only an explicit call to UpdateStatus can update status object, and this is only done by the RadixEnvironment controller.
+		WithOrphaned(false).
```
nilsgstrabo, line 37: No need to set the orphaned property since we never call UpdateStatus. Status should also only be set by the environment controller

pkg/apis/applicationconfig/environment.go
```diff
@@ -0,0 +1,101 @@
...
+func (app *ApplicationConfig) createRadixEnvironment(radixEnvironment *radixv1.RadixEnvironment, logger *log.Entry) error {
+	created, err := app.radixclient.RadixV1().RadixEnvironments().Create(context.Background(), radixEnvironment, metav1.CreateOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to create RadixEnvironment: %v", err)
+	}
+	logger.Debugf("Created RadixEnvironment (revision %s)", created.GetResourceVersion())
+	return nil
+}
+
+// updateRadixEnvironment creates a mergepatch, comparing old and new RadixEnvironments and issues the patch to radix
```
nilsgstrabo, line 66: The comment is wrong. The code performs an Update, not a Patch

pkg/apis/applicationconfig/environment.go
```diff
@@ -0,0 +1,101 @@
...
+func (app *ApplicationConfig) updateRadixEnvironment(oldRE *radixv1.RadixEnvironment, newRE *radixv1.RadixEnvironment, logger *log.Entry) error {
+	radixEnvironment := oldRE.DeepCopy()
+	logger.Debugf("update RadixEnvironment (revision %s)", oldRE.GetResourceVersion())
+	name := newRE.GetName()
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
nilsgstrabo, line 71: When using RetryOnConflict, fn should get the the existing RadixEnvironment. Remove the oldRE parameter and Get() + DeepCopy() at the start of the function

nilsgstrabo (changes requested):
Added some comments

pkg/apis/applicationconfig/environment.go
satr: Fixed

pkg/apis/applicationconfig/environment.go
satr: Fixed

pkg/apis/applicationconfig/environment.go
satr: Fixed

nilsgstrabo (approved):

PR #1043, opened 2024-01-29
fix incorrect adgroups for reader role bindings

nilsgstrabo:
Fixes bug where RR adGroups (admin) was used as subjects for both adm and reader rolebindings for component secrets

### DIFF
--- a/charts/radix-operator/Chart.yaml
+++ b/charts/radix-operator/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.29.2
-appVersion: 1.49.1
+version: 1.29.3
+appVersion: 1.49.3
 kubeVersion: ">=1.24.0"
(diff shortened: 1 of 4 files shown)


charts/radix-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.29.2
-appVersion: 1.49.1
+version: 1.29.3
+appVersion: 1.49.3
```
Richard87, line 4: Bump (app)Version

Richard87 (changes requested):
looks good, but helm chart needs bumping :)

charts/radix-operator/Chart.yaml
nilsgstrabo: fixed

Richard87 (approved):
godkjent

PR #1044, opened 2024-01-31
Retain external dns secret on component change

nilsgstrabo:
Moved creation of external DNS TLS secrets and certificates out of the "sync component" code branch.

The certificate is now created explicitly instead of implicitly via Ingress annotations. 

The certs and secrets are no longer labelled with the component name, but with the FQDN instead, and garbage collection of a component will therefore no longer delete external DNS certs and secrets. External DNS has its own garbage collection routine.

Tested with a full sync of secrets and certs restored from eu-18, and all secrets were successfully relabelled and no data was lost.


Richard87 (approved):
Godkjent

PR #1050, opened 2024-02-08
Add src and image to envs

satr:


pipeline-runner/steps/apply_radixconfig.go
```diff
@@ -30,10 +30,6 @@ import (
 	"sigs.k8s.io/yaml"
 )
```
nilsgstrabo, line 31: What happened to multi component builds? It seems like components with the same src and dockerfileName are no longer built once as a multi component, but instead built separately.

Richard87 (approved):
Godkjent

Richard87 (approved):
godkjent

PR #1052, opened 2024-02-19
add new variable in oauth-aux to support oauth2-proxy v7.6.0

nilsgstrabo:
Ref: https://github.com/equinor/radix-platform/issues/1099


Richard87 (approved):
godkjent

PR #1095, opened 2024-04-22
Use read only fs for Ouath and Job API component

Richard87:


pkg/apis/kube/config_maps.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 
+	_ "github.com/equinor/radix-operator/pkg/apis/test/initlogger"
```
nilsgstrabo, line 8:
```suggestion
```
This is not a `test` file, so the logger should not be imported

pkg/apis/utils/resources/resources.go
```diff
@@ -0,0 +1,72 @@
+package resources
+
+import (
+	"github.com/rs/zerolog/log"
+	corev1 "k8s.io/api/core/v1"
+	resourcev1 "k8s.io/apimachinery/pkg/api/resource"
```
nilsgstrabo, line 6:
```suggestion
	"k8s.io/apimachinery/pkg/api/resource"
```
resource is not versioned

pkg/apis/utils/resources/resources.go
```diff
@@ -0,0 +1,72 @@
...
+		if resources.Requests == nil {
+			resources.Requests = corev1.ResourceList{}
+		}
+
+		resources.Limits[corev1.ResourceMemory] = mem
+		resources.Requests[corev1.ResourceMemory] = mem
+	}
+}
+
+func WithCPU(cpu string) ResourceOption {
```
nilsgstrabo, line 31: Consider using type resource.Quantity instead of string. Then you don't need to parse it

nilsgstrabo (changes requested):
Added a few suggestions. Also, the chart and app version should be updated.

pkg/apis/test/utils.go
```diff
@@ -38,6 +46,23 @@ func NewTestUtils(client kubernetes.Interface, radixclient radixclient.Interface
...
+	logLevel, err := zerolog.ParseLevel(logLevelStr)
+	if err != nil {
+		logLevel = zerolog.InfoLevel
+	}
+
+	zerolog.ErrorStackMarshaler = pkgerrors.MarshalStack
+	zerolog.SetGlobalLevel(logLevel)
+	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: time.Kitchen})
+	zerolog.DefaultContextLogger = &log.Logger
+}
```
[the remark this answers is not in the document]
Richard87: fixed

pkg/apis/utils/resources/resources.go
Richard87: fixed

pkg/apis/utils/resources/resources.go
Richard87: fixed

PR #1102, opened 2024-04-29
return error instead of taking context

Richard87:
Currently fails with 
```
--- FAIL: TestControllerSuite (10.00s)
    --- FAIL: TestControllerSuite/Test_RadixDNSAliasEvents (10.00s)
        controller_test_suite.go:68:
                Error Trace:    /home/richard/projects/radix/radix-operator/radix-operator/common/controller_test_suite.go:68
                                                        /home/richard/projects/radix/radix-operator/radix-operator/dnsalias/controller_test.go:72
                Error:          Timeout waiting for Sync should be called on add RadixDNSAlias
                Test:           TestControllerSuite/Test_RadixDNSAliasEvents
        controller.go:269: missing call(s) to *common.MockHandler.Sync(is anything, is equal to  (string), is equal to alias-alias-1 (string), is equal to &{<nil> false} (*record.FakeRecorder)) /home/richard/projects/radix/radix-operator/radix-operator/dnsalias/controller_test.go:69
        controller.go:269: aborting test due to missing call(s)
FAIL

```


pkg/apis/dnsalias/syncer.go
```diff
@@ -57,28 +58,30 @@ func NewSyncer(kubeClient kubernetes.Interface, kubeUtil *kube.Kube, radixClient
 
 // OnSync is called by the handler when changes are applied and must be
 // reconciled with current state.
-func (s *syncer) OnSync() error {
+func (s *syncer) OnSync(ctx context.Context) error {
+	s.logger = log.Ctx(ctx).With().Str("resource_kind", radixv1.KindRadixDNSAlias).Str("resource_name", cache.MetaObjectToName(&s.radixDNSAlias.ObjectMeta).String()).Logger()
```
nilsgstrabo, line 62: Should we remove this line and use the logger set in NewSyncer(). We should probably decide later how to inject loggers

Makefile
```diff
@@ -149,16 +143,46 @@ radixdnsalias-crd: temp-crds
...
+generate: bootstrap code-gen crds mocks
+
+.PHONY: verify-generate
+verify-generate: bootstrap generate
+	git diff --exit-code
+
+HAS_SWAGGER       := $(shell command -v swagger;)
+HAS_GOLANGCI_LINT := $(shell command -v golangci-lint;)
+HAS_MOCKGEN       := $(shell command -v mockgen;)
+HAS_CONTROLLER_GEN := $(shell command -v controller-gen)
```
nilsgstrabo, line 167:
```suggestion
HAS_CONTROLLER_GEN := $(shell command -v controller-gen;)
```
Missing semicolor. I got error `make: command: Command not found` when running for example `make generate`

nilsgstrabo (changes requested):
Looks good, just a few minor suggestions

nilsgstrabo (approved):

PR #1131, opened 2024-06-14
1101-refine-scheduled-job-batch-statuses

satr:


charts/radix-operator/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: radix-operator
-version: 1.36.0
-appVersion: 1.56.0
+version: 1.36.1
+appVersion: 1.56.1
```
nilsgstrabo, line 4: Update to 1.36.2 and 1.56.2 as *.1 is already used

nilsgstrabo (changes requested):
Update chart version.
Could you also remove one of the two functions isBatchJobPhaseDone and isJobStatusDone as they are identical

nilsgstrabo (approved):

charts/radix-operator/Chart.yaml
satr: fixed

PR #1134, opened 2024-06-19
add-batchstatus-rules

satr:


pkg/apis/radix/v1/radixbatchtypes.go
```diff
@@ -159,6 +160,40 @@ const (
...
+	RadixBatchJobApiStatusWaiting = "Waiting"
+
+	// RadixBatchJobApiStatusStopping job is stopping
+	RadixBatchJobApiStatusStopping = "Stopping"
+
+	// RadixBatchJobApiStatusStopped job stopped
+	RadixBatchJobApiStatusStopped = "Stopped"
+
+	// RadixBatchJobApiStatusDeadlineExceeded job exceeded maximum allowed attempts to run
+	RadixBatchJobApiStatusDeadlineExceeded = "DeadlineExceeded"
```
nilsgstrabo, line 188: This status is not used in existing job or batch statuses in job-scheduler so we should remove it.

satr: Removed

nilsgstrabo (approved):

PR #1152, opened 2024-08-02
Improve pipeline sync

satr:


charts/radix-operator/values.yaml
```diff
@@ -161,6 +161,7 @@ radixGroups:
 
 deploymentsPerEnvironmentHistoryLimit: 10
 pipelineJobsHistoryLimit: 5
+pipelineJobsHistoryPeriodLimit: "30d"
```
nilsgstrabo, line 164: "d" (day?) is not a valid Go duration string. "h" (hour) is the largest, so the default value must be changed to "720h"

radix-operator/job/handler.go
```diff
@@ -25,31 +26,43 @@ const (
...
 
-// Handler Instance variables
-type Handler struct {
+// Handler Common handler interface
+type Handler interface {
+	// Sync Is created on sync of resource
+	Sync(ctx context.Context, namespace, name string, eventRecorder record.EventRecorder) error
+	// CleanupJobHistory Cleanup the pipeline job history
+	CleanupJobHistory(ctx context.Context, radixJob *v1.RadixJob)
+}
```
nilsgstrabo, line 35:
```suggestion
type Handler interface {
	common.Handler
	// CleanupJobHistory Cleanup the pipeline job history
	CleanupJobHistory(ctx context.Context, radixJob *v1.RadixJob)
}
```
"Inherit" the common.Handler instead of redecalring the Sync method

pkg/apis/job/job_history.go
```diff
@@ -4,107 +4,185 @@ import (
...
+	}
+}
+
+// Cleanup the pipeline job history
+func (h *history) Cleanup(ctx context.Context, appName, radixJobName string) {
+	namespace := utils.GetAppNamespace(appName)
+	if _, ok := h.namespacesRequestsToCleanup.LoadOrStore(namespace, struct{}{}); ok {
+		return // a request to clean up history in this namespace already exists
+	}
+	go func() {
```
nilsgstrabo, line 67: I suggest to not use Go routine and channel, as we discussed.

radix-operator/job/handler.go
satr: Fixed

charts/radix-operator/values.yaml
satr: Fixed

nilsgstrabo (approved):

nilsgstrabo (approved):

PR #1169, opened 2024-08-19
1159 bug git ssh keys secret unexpectedly regenerated

nilsgstrabo:


pkg/apis/application/secrets.go
```diff
@@ -15,97 +14,143 @@ import (
...
+		knownHostsSecretKey:             knownHostsSecret.Data[knownHostsSecretKey],
+	}
+
+	return original, desired, derivedPublicKey, nil
 }
 
-func deriveKeyPairFromSecret(secret *corev1.Secret) (*utils.DeployKey, error) {
-	privateKey := string(secret.Data[defaults.GitPrivateKeySecretKey])
-	deployKey, err := utils.DeriveDeployKeyFromPrivateKey(privateKey)
+func (app *Application) getCurrentAndDesiredGitPublicDeployKeyConfigMap(ctx context.Context, publicKey string) (original, desired *corev1.ConfigMap, err error) {
```
satr, line 127:
```suggestion
func (app *Application) getCurrentAndDesiredGitPublicDeployKeyConfigMap(ctx context.Context, publicKey string) (*corev1.ConfigMap, *corev1.ConfigMap, err error) {
```

satr: It maybe confusing having variables both return `return original, desired, nil` and method interface

Richard87 (approved):
lgtm!

satr (approved):

PR #1198, opened 2024-09-23
add support for filtering public ingress traffic by IP or CIDR

nilsgstrabo:


pkg/apis/ingress/ingressannotationprovider.go
```diff
@@ -128,3 +128,30 @@ func (provider *oauth2AnnotationProvider) GetAnnotations(component radixv1.Radix
...
+
+	for i, addr := range *component.GetNetwork().Ingress.Public.Allow {
+		if i > 0 {
+			sb.WriteString(",")
+		}
+		sb.WriteString(string(addr))
+	}
+	annotations["nginx.ingress.kubernetes.io/whitelist-source-range"] = sb.String()
+
+	return annotations, nil
```
satr, line 156: Optional suggestion
```suggestion
	addressList := slice.Reduce(*component.GetNetwork().Ingress.Public.Allow, []string{}, func(acc []string, addr radixv1.IPOrCIDR) []string {
		return append(acc, string(addr))
	})
	return map[string]string{"nginx.ingress.kubernetes.io/whitelist-source-range": strings.Join(addressList, ",")}, nil
```

satr (approved):

pkg/apis/ingress/ingressannotationprovider.go
```diff
@@ -128,3 +128,30 @@ func (provider *oauth2AnnotationProvider) GetAnnotations(component radixv1.Radix
...
+
+// NewIngressPublicAllowListAnnotationProvider provides Ingress annotations for allowing
+// only public traffic from IP adresses defined in Network.Ingress.Public.Allow field
+func NewIngressPublicAllowListAnnotationProvider() AnnotationProvider {
+	return &ingressPublicAllowListAnnotationProvider{}
+}
+
+type ingressPublicAllowListAnnotationProvider struct{}
+
+func (*ingressPublicAllowListAnnotationProvider) GetAnnotations(component radixv1.RadixCommonDeployComponent, _ string) (map[string]string, error) {
```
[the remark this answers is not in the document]
nilsgstrabo: fixed

satr (dismissed):

pkg/apis/ingress/ingressannotationprovider.go
nilsgstrabo: Fixed, but used slice.Map instead of Reduce

satr (approved):

PR #1228, opened 2024-11-21
Fix restart batchjob infinite loop

nilsgstrabo:
A bug in RadixBatch status update function caused jobStatus.Phase to be set to empty string. Since this is not an allowed values, the update failed causing the RB to be resynced over and over because the jobStatus.Restart was not updated either.

### DIFF
--- a/pkg/apis/batch/syncer.go
+++ b/pkg/apis/batch/syncer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/equinor/radix-common/utils/slice"
 	"github.com/equinor/radix-operator/pkg/apis/config"
 	"github.com/equinor/radix-operator/pkg/apis/kube"
 	radixv1 "github.com/equinor/radix-operator/pkg/apis/radix/v1"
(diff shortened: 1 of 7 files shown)


pkg/apis/batch/syncer.go
```diff
@@ -83,7 +88,7 @@ func (s *syncer) reconcile(ctx context.Context) error {
 			return fmt.Errorf("batchjob %s: failed to reconcile kubejob: %w", batchJob.Name, err)
 		}
 
-		if i%syncStatusForEveryNumberOfBatchJobsReconciled == 0 {
+		if i%syncStatusForEveryNumberOfBatchJobsReconciled == (syncStatusForEveryNumberOfBatchJobsReconciled - 1) {
```
Richard87, line 91: Why are we only syncing after 10 jobs, and not on every iteration? And what happens if a job has less than 10, eg. 5, when will they get synced?

Richard87 (approved):
lgtm

pkg/apis/batch/syncer.go
nilsgstrabo: The batch controller will always perform a status update as the final step in OnSync: https://github.com/equinor/radix-operator/blob/3a31375212221311e9f7742c7b80e6917749ea14/pkg/apis/batch/syncer.go#L63

The reason we update status for every 10 jobs is to give some kind of feedback about the sync progress when a batch contains many jobe. If a batch has 100-200 jobs, it can easily take more than a minute to iterate and process them all, and instead of updating status at the end, we update it for every 10 jobs to give status feedback faster just to let the user know that "things are in progress"

PR #1234, opened 2024-12-02
feat: Add Healtchchecks to RadixConfig

Richard87:
- [x] Validate only of probe handler is defined
- [x] Test data is copied correctly


pkg/apis/radix/v1/radixhealthchecktypes.go
```diff
@@ -0,0 +1,209 @@
...
+	// Must be 1 for liveness and startup.
+	// +kubebuilder:validation:Minimum=1
+	// +default=1
+	// +optional
+	SuccessThreshold *int32 `json:"successThreshold,omitempty"`
+	// Minimum consecutive failures for the probe to be considered failed after having succeeded.
+	// +kubebuilder:validation:Minimum=1
+	// +default=3
+	// +optional
+	FailureThreshold *int32 `json:"failureThreshold,omitempty"`
```
nilsgstrabo, line 39: Why not use the same data type and non-pointer as in corev1.Probe? I don't think we need to define defaults since this is something Kubernetes will handle.

pkg/apis/radix/v1/radixhealthchecktypes.go
```diff
@@ -0,0 +1,209 @@
...
+	// // Set this value longer than the expected cleanup time for your process.
+	// // If this value is nil, the pod's terminationGracePeriodSeconds will be used. Otherwise, this
+	// // value overrides the value provided by the pod spec.
+	// // Value must be non-negative integer. The value zero indicates stop immediately via
+	// // the kill signal (no opportunity to shut down).
+	// // This is a beta field and requires enabling ProbeTerminationGracePeriod feature gate.
+	// // +kubebuilder:validation:Minimum=1
+	// // +default=30
+	// // +optional
+	// TerminationGracePeriodSeconds *int64 `json:"terminationGracePeriodSeconds,omitempty"`
```
nilsgstrabo, line 53:
```suggestion
```
Suggest we remove this if we don't intend to use it

pkg/apis/radix/v1/radixhealthchecktypes.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	return &corev1.Probe{
+		ProbeHandler:        rp.RadixProbeHandler.MapToCoreProbe(),
+		InitialDelaySeconds: pointers.Val(rp.InitialDelaySeconds),
+		TimeoutSeconds:      pointers.Val(rp.TimeoutSeconds),
+		PeriodSeconds:       pointers.Val(rp.PeriodSeconds),
+		SuccessThreshold:    pointers.Val(rp.SuccessThreshold),
+		FailureThreshold:    pointers.Val(rp.FailureThreshold),
+		// TerminationGracePeriodSeconds: rp.TerminationGracePeriodSeconds,
```
nilsgstrabo, line 68:
```suggestion
```
Suggest we remove this if we don't intend to use it

pkg/apis/radix/v1/radixhealthchecktypes.go
```diff
@@ -0,0 +1,209 @@
...
+	return corev1.ProbeHandler{
+		Exec:      p.Exec.MapToCoreProbe(),
+		HTTPGet:   p.HTTPGet.MapToCoreProbe(),
+		TCPSocket: p.TCPSocket.MapToCoreProbe(),
+		GRPC:      p.GRPC.MapToCoreProbe(),
+	}
+}
+
+// RadixProbeHTTPGetAction describes an action based on HTTP Get requests.
+type RadixProbeHTTPGetAction struct {
```
nilsgstrabo, line 95: Shoudl we use the same data types for fields as in the corev1 struct? Goes for the other probe actions as well

nilsgstrabo (changes requested):
Added a few suggestions that we should discuss

pkg/apis/radix/v1/radixhealthchecktypes.go
Richard87: added todo

pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -805,6 +815,28 @@ type RadixJobComponentPayload struct {
 	Path string `json:"path"`
 }
 
+type RadixHealthChecks struct {
```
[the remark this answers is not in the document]
Richard87: fixed

pkg/apis/radix/v1/radixhealthchecktypes.go
Richard87: agreed, fixed

pkg/apis/radix/v1/radixhealthchecktypes.go
Richard87: agreed

nilsgstrabo (approved):

PR #1278, opened 2025-02-04
Managed identity for volume mount

satr:


pkg/apis/internal/identity.go
```diff
@@ -0,0 +1,11 @@
+package internal
+
+import "github.com/equinor/radix-operator/pkg/apis/radix/v1"
+
+// GetIdentityClientId Get the client id of the identity
+func GetIdentityClientId(identity *v1.Identity) string {
```
nilsgstrabo, line 6: Perhaps we could supprt getting the clientId for identity.azure by adding the following receiver functions for Identity and AzureIdentity structs in radixapptypes.go. It also handles nil references

```
func (i *Identity) GetAzure() *AzureIdentity {
	if i == nil {
		return nil
	}
	return i.Azure
}

func (az *AzureIdentity) GetClientId() string {
	if az == nil {
		return ""
	}
	return az.ClientId
}
```

Then we can can call these methods on any *Identity objects
```
var i *radixv1.Identity

clientId:=identity.GetAzure().GetClientId()
```

pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -1464,21 +1466,25 @@ func hasNonResourceTypeTriggers(config *radixv1.RadixHorizontalScaling) bool {
...
 	var errs []error
 	for _, component := range app.Spec.Components {
-		if err := validateVolumeMounts(component.VolumeMounts); err != nil {
+		hasComponentIdentityAzureClientId := len(internal.GetIdentityClientId(component.Identity)) > 0
+		if err := validateVolumeMounts(component.VolumeMounts, hasComponentIdentityAzureClientId); err != nil {
 			errs = append(errs, volumeMountValidationFailedForComponent(component.Name, err))
 		}
 		for _, envConfig := range component.EnvironmentConfig {
-			if err := validateVolumeMounts(envConfig.VolumeMounts); err != nil {
+			hasEnvIdentityAzureClientId := len(internal.GetIdentityClientId(envConfig.GetIdentity())) > 0
```
nilsgstrabo, line 1474: It is not required to define identity on environment if it is defined on component level. Validation must handle these situations.

pkg/apis/volumemount/volumemount.go
```diff
@@ -541,18 +546,16 @@ func getVolumeHandle(namespace, componentName, pvName, storageName string) strin
...
+	return UseAzureIdentityForVolumeMount(radixVolumeMount)
 }
 
-func getIdentityClientId(identity *radixv1.Identity) string {
-	if identity != nil && identity.Azure != nil && len(identity.Azure.ClientId) > 0 {
-		return identity.Azure.ClientId
-	}
-	return ""
+// UseAzureIdentityForVolumeMount Returns true if the volume mount should use Azure Identity
+func UseAzureIdentityForVolumeMount(radixVolumeMount *radixv1.RadixVolumeMount) bool {
```
nilsgstrabo, line 556: This function should be private since it is only used internally

pkg/apis/volumemount/volumemount.go
```diff
@@ -541,18 +546,16 @@ func getVolumeHandle(namespace, componentName, pvName, storageName string) strin
 }
 
 func getUseAzureIdentity(identity *radixv1.Identity, radixVolumeMount *radixv1.RadixVolumeMount) bool {
-	if len(getIdentityClientId(identity)) == 0 {
+	if len(internal.GetIdentityClientId(identity)) == 0 {
```
nilsgstrabo, line 549: Why do we need to check if Identity.Azure.ClientId is set? If the user has set useAzureIdentity on the volumeMount, but not configured Identity.Azure, then this will be picked up by RA validation.
I think we should just set the PV attributes for workload identity, even if this means an empty value for clientid (for example if there is a bug in RA validation), which would obviously cause mounting error.

pkg/apis/volumemount/volumemount.go
```diff
@@ -592,9 +595,6 @@ func getCsiAzurePvAttributes(namespace string, radixVolumeMount *radixv1.RadixVo
 			}
 		}
 	}
```
nilsgstrabo, line 597: Remove unused parameters pvName and pvcName

satr: Removed

pkg/apis/volumemount/volumemount.go
```diff
@@ -122,10 +123,12 @@ func GarbageCollectCsiAzureVolumeResourcesForDeployComponent(ctx context.Context
 func CreateOrUpdateVolumeMountSecrets(ctx context.Context, kubeUtil *kube.Kube, appName, namespace, componentName string, volumeMounts []radixv1.RadixVolumeMount) ([]string, error) {
 	var volumeMountSecretsToManage []string
 	for _, volumeMount := range volumeMounts {
+		if UseAzureIdentityForVolumeMount(&volumeMount) {
```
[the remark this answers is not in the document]
satr: Fixed

pkg/apis/volumemount/volumemount.go
satr: Removed check

pkg/apis/volumemount/volumemount.go
satr: Fixed

pkg/apis/radixvalidators/validate_ra.go
```diff
@@ -1464,21 +1466,25 @@ func hasNonResourceTypeTriggers(config *radixv1.RadixHorizontalScaling) bool {
...
 	}
 	for _, job := range app.Spec.Jobs {
-		if err := validateVolumeMounts(job.VolumeMounts); err != nil {
+		hasJobIdentityAzureClientId := len(internal.GetIdentityClientId(job.Identity)) > 0
+		if err := validateVolumeMounts(job.VolumeMounts, hasJobIdentityAzureClientId); err != nil {
 			errs = append(errs, volumeMountValidationFailedForJobComponent(job.Name, err))
 		}
 		for _, envConfig := range job.EnvironmentConfig {
-			if err := validateVolumeMounts(envConfig.VolumeMounts); err != nil {
+			hasEnvIdentityAzureClientId := len(internal.GetIdentityClientId(envConfig.GetIdentity())) > 0
```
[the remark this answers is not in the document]
satr: Fixed with check `hasComponentIdentityAzureClientId || hasEnvIdentityAzureClientId`

pkg/apis/radixvalidators/validate_ra.go
satr: Fixed with check hasComponentIdentityAzureClientId || hasEnvIdentityAzureClientId

pkg/apis/volumemount/volumemount.go
```diff
@@ -564,7 +553,7 @@ func getCsiAzurePvLabels(appName, namespace, componentName string, radixVolumeMo
 	}
 }
 
-func getCsiAzurePvAttributes(namespace string, radixVolumeMount *radixv1.RadixVolumeMount, pvName, pvcName string, useAzureIdentity bool, clientId string) map[string]string {
+func getCsiAzurePvAttributes(namespace string, radixVolumeMount *radixv1.RadixVolumeMount, useAzureIdentity bool, clientId string) map[string]string {
```
nilsgstrabo, line 556: No need to pass useAzureIdentity and clientId as parameters, since radixVolumeMount is already passed and the function get get the values from this object

nilsgstrabo (approved):
Approved with some optional suggestions

pkg/apis/volumemount/test.go
```diff
@@ -499,7 +531,6 @@ func getMountOptionsInRandomOrder(props expectedPvcPvProperties, extraOptions ..
 
 func setVolumeMountAttribute(pv *v1.PersistentVolume, radixVolumeMountType radixv1.MountType, containerName, pvcName string) {
```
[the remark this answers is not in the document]
satr: Fixed

pkg/apis/volumemount/volumemount.go
satr: Fixed

nilsgstrabo (approved):

PR #1286, opened 2025-02-13
Identity for oauth aux

satr:


pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -1378,6 +1378,10 @@ type OAuth2 struct {
 	// Settings for Redis store when SessionStoreType is redis.
 	// +optional
 	RedisStore *OAuth2RedisStore `json:"redisStore,omitempty"`
+
+	// UseAzureIdentity defines that credentials for authenticating using Azure Workload Identity instead of using a ClientSecret.
+	// +optional
+	UseAzureIdentity *bool `json:"useAzureIdentity,omitempty"`
```
Richard87, line 1384: To make it slighly more future proof, could we use something like `ClientCredentials: "secret"|"AzureWorkloadIdentity"` instead? (in case oauth2proxy changes their implementation and supports more providers that we want to support)

Richard87 (changes requested):
Is it valuable to refactor, and move the oauth2proxy related code to its own package (`pkg/apis/oauth2proxy` or similar)?

satr:
> Is it valuable to refactor, and move the oauth2proxy related code to its own package (`pkg/apis/oauth2proxy` or similar)?

Moving to oauth2proxy requires moving other code due to cyclic references and several private methods. I think this need to be done without reviewving new functionality.

pkg/apis/radix/v1/radixapptypes.go
satr: Replaced

Richard87 (approved):
Looks good to me!

PR #1311, opened 2025-03-11
1290 changes to storage account files are not updated

nilsgstrabo:
- Updates to blobfuse2 volumemount:
  - add support for block cache. This is the default unless streaming.enabled=false, which will use file caching as before
  - added timeout options for file and attribute cache
- split PV spec generation for blobfuse2 and the deprecated volume type into two structs
- misc refactoring and test cleanup
- removed unused SkuName and BinidngMode from VolumeMount and BlobFuse2 structs
- added a performance fix to radixbatch syncer to prevent unneccessary calls generated volumemounts unless an actual k8s job must be created.


pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -1117,10 +1182,79 @@ type RadixBlobFuse2VolumeMount struct {
...
+	// +optional
+	DiskSize *uint32 `json:"diskSize,omitempty"`
+
+	// Timeout (in seconds) for which persisted data remains in disk cache. Default 120.
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	DiskTimeout *uint32 `json:"diskTimeout,omitempty"`
+
+	// Max number of blocks to prefetch. Default 11.
+	// Value must be 0 (prefetcing disabled) or greater or equal to 11.
```
satr, line 1204:
```suggestion
	// Value must be 0 (prefetching disabled) or greater or equal to 11.
```

pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -1091,10 +1135,31 @@ type RadixBlobFuse2VolumeMount struct {
...
+	// +optional
+	CacheMode *BlobFuse2CacheMode `json:"cacheMode,omitempty"`
+
+	// Configure file cache settings.
+	// +optional
+	FileCacheOptions *BlobFuse2FileCacheOptions `json:"fileCache,omitempty"`
+
+	// Configure block cache settings.
+	// +optional
+	BlockCacheOptions *BlobFuse2BlockCacheOptions `json:"blockCache,omitempty"`
```
satr, line 1131: Suggestion to describe which option is applicable for each `CacheMode`

pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -1117,10 +1182,79 @@ type RadixBlobFuse2VolumeMount struct {
...
+
+const (
+	BlobFuse2CacheModeFile     BlobFuse2CacheMode = "File"
+	BlobFuse2CacheModeBlock    BlobFuse2CacheMode = "Block"
+	BlobFuse2CacheModeDirectIO BlobFuse2CacheMode = "DirectIO"
+)
+
+// BlobFuse2AttributeCacheOptions defines options for attribute cache
+type BlobFuse2AttributeCacheOptions struct {
+	// The timeout for the attribute cache entries. Default 0.
```
satr, line 1171:
```suggestion
	// The timeout in seconds for the attribute cache entries. Default 0.
```

pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -1117,10 +1182,79 @@ type RadixBlobFuse2VolumeMount struct {
...
+
+	// Number of worker thread responsible for upload/download jobs. Default 8.
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	Parallelism *uint32 `json:"parallelism,omitempty"`
+}
+
+// BlobFuse2FileCacheOptions defines options for file cache
+type BlobFuse2FileCacheOptions struct {
+	// The timeout for which file cache is valid. Default 120.
```
satr, line 1221:
```suggestion
	// The timeout (in seconds) for which file cache is valid. Default 120.
```

pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -1117,10 +1182,79 @@ type RadixBlobFuse2VolumeMount struct {
...
+type BlobFuse2BlockCacheOptions struct {
+	// Size (in MB) of a block to be downloaded as a unit. Default is 4.
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	BlockSize *uint32 `json:"blockSize,omitempty"`
+
+	// Size (in MB) of total memory preallocated for block-cache.
+	// Minimum value:
+	// - if prefetchCount > 0: prefetchCount * blockSize
+	// - if prefetchCount = 0: blockSize
```
satr, line 1187: How is it related to `PoolSize`? Or it should be part of the comment for the `PrefetchCount` property?

pkg/apis/radix/v1/radixapptypes.go
satr, line 1204: Can it be described simplier - like in the error: "... or greater than 10"

pkg/apis/volumemount/defaults.go
```diff
@@ -5,16 +5,10 @@ const (
...
 
 	csiPersistentVolumeClaimNameTemplate = "pvc-%s-%s"              // pvc-<volumename>-<randomstring5>
 	csiPersistentVolumeNameTemplate      = "pv-radixvolumemount-%s" // pv-<guid>
 
-	csiMountOptionGid     = "gid"      // Volume mount owner GroupID. Used when drivers do not honor fsGroup securityContext setting
-	csiMountOptionUid     = "uid"      // Volume mount owner UserID. Used instead of GroupID
-	csiMountOptionUseAdls = "use-adls" // Use ADLS or Block Blob
-
 	csiVolumeAttributeProtocolParameterFuse  = "fuse"  // Protocol "blobfuse"
 	csiVolumeAttributeProtocolParameterFuse2 = "fuse2" // Protocol "blobfuse2"
```
satr, line 13: Not longer in use

nilsgstrabo: removed

pkg/apis/radix/v1/radixapptypes.go
nilsgstrabo: fixed

pkg/apis/radix/v1/radixapptypes.go
nilsgstrabo: When the driver reads a block from the storage account it must be stored somewhere. This "somewhere" is in the memory pool, which means that the pool size must be large enough to hold these blocks. When prefetching=0 (disabled) then the driver will read one block. The poolsize must therefore be minimun the size of one block. When prefetching is set, then the poolsize must be at least the size of all blocks that are prefetched

pkg/apis/radix/v1/radixapptypes.go
nilsgstrabo: updated

satr (approved):

PR #1339, opened 2025-04-23
1305 rewrite applicationconfig apply privateimagehubsecret to sync secret not recreate them

nilsgstrabo:
- Use CreateSecret/UpdateSecret instead of deprecated ApplySecret function to prevent overwrite of external annotations
- Update tests to include checking that password is retained between syncs
- Renamed the two functions that converts from/to docker auth config from secret data. These functions are used in radix-api, and I will update them there once radix-operator is merged and tagged


Richard87 (approved):
Just one question, but looks good! :)

PR #1354, opened 2025-05-07
1205 refresh build kit cache

satr:


pipeline-runner/steps/applyconfig/step.go
```diff
@@ -39,53 +41,94 @@ func NewApplyConfigStep() model.Step {
...
+
+func (step *ApplyConfigStepImplementation) applyRadixApplication(ctx context.Context, pipelineInfo *model.PipelineInfo) error {
+	ra := pipelineInfo.RadixApplication
+	namespace := operatorutils.GetAppNamespace(ra.GetName())
+	dnsAliasConfig := pipelineInfo.PipelineArguments.DNSConfig
+
+	existingRA, err := step.GetRadixClient().RadixV1().RadixApplications(namespace).Get(ctx, ra.Name, metav1.GetOptions{})
+	if err != nil {
+		if kubeerrors.IsNotFound(err) {
+			return step.createRadixApplication(ctx, namespace, ra, dnsAliasConfig)
```
nilsgstrabo, line 87: I think it we could inline the code in createRadixApplication here (and delete createRadixApplication). All it does is perform validation and a call to Create

Also we should move the call to radixvalidators.CanRadixApplicationBeInserted to the beginning of this function

pipeline-runner/steps/applyconfig/step.go
```diff
@@ -39,53 +41,94 @@ func NewApplyConfigStep() model.Step {
...
 // Run Override of default step method
-func (cli *ApplyConfigStepImplementation) Run(ctx context.Context, pipelineInfo *model.PipelineInfo) error {
+func (step *ApplyConfigStepImplementation) Run(ctx context.Context, pipelineInfo *model.PipelineInfo) error {
 	printPrepareBuildContext(ctx, pipelineInfo.BuildContext)
 
-	applicationConfig := application.NewApplicationConfig(cli.GetKubeClient(), cli.GetKubeUtil(),
-		cli.GetRadixClient(), cli.GetRegistration(), pipelineInfo.RadixApplication,
-		pipelineInfo.PipelineArguments.DNSConfig)
-
-	pipelineInfo.SetApplicationConfig(applicationConfig)
```
nilsgstrabo, line 64: pipelineInfo.SetApplicationConfig method can be deleted as it is no longer used, and so can the applicationConfig.GetRadixApplicationConfig method that it calls

nilsgstrabo (changes requested):
Added some suggestion to cleanup unused code, and "simplyfy" the applyRadixApplication method in applyconfig step

pipeline-runner/steps/applyconfig/step.go
satr: Fixed

pipeline-runner/steps/applyconfig/step.go
satr: Moved

nilsgstrabo (approved):

Issue #1351, opened 2025-05-07
Unable to promote RD when component has {imageTagName} and target environment does not have imageTagName prop


nilsgstrabo:
pipeline-runner should not validate the the source environment has imageTagName set when running promote jobs.

Logged message in pipeline-runner
```
2025-05-07T11:56:30Z ERR Failed to apply config for application oauth-demo2. Error: component redis2 in environment qa: missing required imageTagName in a component, an environmentConfig or in a pipeline argument
```

PR #1363, opened 2025-05-21
fix: unable to promote RD when imageTagName is missing in target environment

nilsgstrabo:
This was previously fixed in https://github.com/equinor/radix-operator/issues/1351, but due to missing tests, the changes were reverted/reintroduced in https://github.com/equinor/radix-operator/pull/1354


Richard87 (approved):
godkjent

PR #1383, opened 2025-06-16
Configure redis in radixconfig

satr:


pkg/apis/radix/v1/radixapptypes.go
```diff
@@ -2256,9 +2266,36 @@ func (oauth2 *OAuth2) GetSessionStoreType() SessionStoreType {
 	if oauth2 == nil {
 		return SessionStoreCookie
 	}
+	if oauth2.SessionStoreTypeIsSystemManaged() {
```
nilsgstrabo, line 2269: GetSessionStoreType() should return the real value of SessionStoreType. It should be up to the consumer (oauth2 aux manager) to decide/translate it into usable values

pkg/apis/deployment/oauthproxyresourcemanager.go
```diff
@@ -483,10 +483,21 @@ func (o *oauthProxyResourceManager) createOrUpdateSecret(ctx context.Context, co
...
-	oauthutil.MergeAuxComponentResourceLabels(secret, o.rd.Spec.AppName, component)
-	if _, ok := secret.Data[defaults.OAuthRedisPasswordKeyName]; ok && component.GetAuthentication().GetOAuth2().GetSessionStoreType() != v1.SessionStoreRedis {
+	oauthutil.MergeAuxOAuthProxyComponentResourceLabels(secret, o.rd.Spec.AppName, component)
+
+	redisPassword, redisPasswordExists := secret.Data[defaults.OAuthRedisPasswordKeyName]
+	if redisPasswordExists && !component.GetAuthentication().GetOAuth2().SessionStoreTypeIsRedis() {
 		delete(secret.Data, defaults.OAuthRedisPasswordKeyName)
 	}
+	if component.GetAuthentication().GetOAuth2().SessionStoreTypeIsSystemManaged() &&
+		(!redisPasswordExists || len(redisPassword) <= 2) {
```
nilsgstrabo, line 493: Why len(redisPassword) <= 2 and not len(redisPassword) == 0?

pkg/apis/radix/v1/radixapptypes.go
satr: Fixed

pkg/apis/deployment/oauthproxyresourcemanager.go
satr: To avoid empty of "xx" secrets

nilsgstrabo (approved):

PR #1397, opened 2025-07-14
extend-scheduled-job-description

satr:
Allow env var as job payload https://github.com/equinor/radix/issues/484
Be able to set imageName in addition to imageTagName when starting jobs https://github.com/equinor/radix/issues/485

### DIFF
--- a/pkg/apis/batch/status.go
+++ b/pkg/apis/batch/status.go
@@ -184,18 +184,17 @@ func (s *syncer) updateJobAndPodStatuses(ctx context.Context, batchJobName strin
 	jobComponentName := s.radixBatch.GetLabels()[kube.RadixComponentLabel]
 	podStatusMap := getPodStatusMap(jobStatus)
 	for _, pod := range s.getJobPods(ctx, batchJobName) {
-		imageInPodSpec := getImageInPodSpec(jobComponentName, pod)
-		podStatus := getOrCreatePodStatusForPod(&pod, jobStatus, podStatusMap)
+		podStatus := getOrCreatePodStatusForPod(&pod, jobStatus, podStatusMap, jobComponentName)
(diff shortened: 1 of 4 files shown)


pkg/apis/radix/v1/radixbatchtypes.go
```diff
@@ -346,6 +356,11 @@ type RadixBatchJobPodStatus struct {
 	// +optional
 	Image string `json:"image"`
 
+	// The name of container image that is specified in PodSpec.
+	// More info: https://kubernetes.io/docs/concepts/containers/images.
+	// +optional
+	ImageInSpec string `json:"imageInSpec"`
```
nilsgstrabo, line 362: Remove this field, and instead use the existing Image field to store the image from the container step instead of the image from containerStatus.

pkg/apis/batch/status.go
```diff
@@ -184,15 +184,23 @@ func (s *syncer) updateJobAndPodStatuses(ctx context.Context, batchJobName strin
 	jobComponentName := s.radixBatch.GetLabels()[kube.RadixComponentLabel]
 	podStatusMap := getPodStatusMap(jobStatus)
 	for _, pod := range s.getJobPods(ctx, batchJobName) {
+		imageInPodSpec := getImageInPodSpec(jobComponentName, pod)
```
nilsgstrabo, line 187: Move this into getOrCreatePodStatusForPod(...), and set RadixBatchJobStatus.Image to the value it returns.
Also remove setting RadixBatchJobStatus.Image from setPodStatusByPodLastContainerStatus(...)

nilsgstrabo (changes requested):
Remove the new `ImageInSpec` field and just use existing `Image` to store the specified image from the pod spec

pkg/apis/radix/v1/radixbatchtypes.go
satr: Removed

PR #1399, opened 2025-07-16
Make authentication in servicebus trigger optional

satr:


pkg/apis/radix/v1/radixhorizontalscalingtypes.go
```diff
@@ -167,6 +167,7 @@ type RadixHorizontalScalingAzureServiceBusTrigger struct {
 	ActivationMessageCount *int `json:"activationMessageCount,omitempty"`
 
 	// Azure Service Bus requires Workload Identity configured with a ClientID
+	// +optional
 	Authentication RadixHorizontalScalingAuthentication `json:"authentication"`
```
nilsgstrabo, line 171:
```suggestion
	Authentication *RadixHorizontalScalingAuthentication `json:"authentication,omitempty"`
```
This should probably be a pointer and have omitempty (to prevent seeing "authentication: null" when marshaling)

satr: fixed - set omitempty, validation already exists

nilsgstrabo (approved):

PR #1518, opened 2025-11-11
feat: extend CRD status with with reconcile state and error

nilsgstrabo:
This PR standardizes status fields across all Radix Custom Resource Definitions (CRDs) to improve reconciliation tracking and error reporting. The changes follow a consistent pattern of adding reconcile state information (reconciled timestamp, observed generation, reconcile status, and error messages) to all CRDs.

Key changes include:

- Standardized status fields across all CRDs with Reconciled, ObservedGeneration, ReconcileStatus, and Message fields
- Moved RetryOnConflict from individual reconcilers to the controller level
- Replaced AddRateLimited() with Add() in controller Enqueue methods to prevent delayed reconciles
- Removed hasSynced callbacks from controllers and refactored tests to use MockHandler
- Generate open-API schema for RadixJob and RadixEnvironment resources


operator/alert/controller.go
```diff
@@ -89,9 +89,6 @@ func NewController(ctx context.Context,
 		UpdateFunc: func(oldObj, newObj interface{}) {
 			newRr := newObj.(*radixv1.RadixRegistration)
 			oldRr := oldObj.(*radixv1.RadixRegistration)
```
Richard87: Why remove this check?

PR #1601, opened 2026-02-11
fix: restrict environment names

Richard87:


webhook/validation/radixapplication/validate.go
```diff
@@ -343,6 +347,38 @@ func envNameValidator(ctx context.Context, ra *radixv1.RadixApplication) ([]stri
...
+		for _, env := range ra.Spec.Environments {
+			envNs := utils.GetEnvironmentNamespace(ra.Name, env.Name)
+
+			// Returns true if namespace exist, but does not belong to the application
+			unavailableNs := slices.ContainsFunc(existingNamespaces.Items, func(ns corev1.Namespace) bool {
+				if ns.Name != envNs {
+					return false
+				}
+				// Namespace with the same name exists, check if it has the correct ownership label
+				if val, exist := ns.Labels[kube.RadixAppLabel]; !exist || val != ra.Name {
```
nilsgstrabo: We don't need exist since we always check for a non-empty string

nilsgstrabo (approved):
Approved, with some optional suggestions

PR #1643, opened 2026-03-17
fix: update git clone command to handle submodules correctly

### LINKED ISSUE #1624: Test and fix git submodules feature
Look into our current tests and make sure we properly test submodules.

Make sure we cchange directory to the cloned repo before running git init.
https://github.com/equinor/radix-operator/blob/master/pkg/apis/git/clone.go#L44

```
commands = append(commands, fmt.Sprintf("git clone %s %s %s && (git submodule update --init --recursive || echo \"Warning: Unable to clone submodules, proceeding without them\")", sshURL, strings.Join(cloneArgs, " "), directory))
```

nilsgstrabo:
This pull request makes a small but important change to the `CloneInitContainersWithContainerName` function in `pkg/apis/git/clone.go` to ensure that submodules are updated in the correct directory after cloning a git repository.

- Bug fix: The `git submodule update --init --recursive` command is now executed inside the cloned repository directory by adding a `cd` into the directory before running the submodule command. This prevents submodule updates from failing due to being run in the wrong location.


Richard87 (approved):
Looks good to me

PR #1749, opened 2026-05-28
fix(tests): add missing env variable and add tests to verify configs are set 

### LINKED ISSUE #1748: Bug: Radixx-operator deoesnt set required env-variable Cluster Name
The Cluster Name is required, but its lost since its not available for radix-operators environment variables

herda1:
TODO:

- [x] Flatten config structure, test if structs can be simplified (avoid "ignore" in annotation)
- [x] Migrate pkg/api/config to operator/config


charts/radix-operator/templates/deployment.yaml
```diff
@@ -39,6 +39,8 @@ spec:
               value: {{ .Values.dnsZone }}
             - name: RADIX_ZONE
               value: {{ .Values.radixZone }}
+            - name: RADIX_CLUSTERNAME
+              value: {{ .Values.clusterName }}
```
herda1, line 43: the actual fix for the failing env

nilsgstrabo (dismissed):

nilsgstrabo (approved):

PR #1781, opened 2026-06-16
fix(radix-job): add resolved commit id and store it

### LINKED ISSUE #1750: Store resolved pipeline commitid in RadixJob status
When a user runs a build or build-deploy pipeline job, the pipeline runner will resolve the the latest commitId for the branch being built, if the commitId was not passed to the pipeline (defined in [RadixJob.Spec.Build.CommitID](https://github.com/equinor/radix-operator/blob/72214eba5cba486f5574d64f83b30e3037da7eb4/pkg/apis/radix/v1/radixjobtypes.go#L210)). This value is currently not stored anywhere on the pipeline job, which makes it difficult for users to know which commit was actually built. Knowing the commitId is also important in a disaster recovery scenario where the ACR is lost and we have to rerun build-deploy jobs to rebuild container images.

The [resolved commit](https://github.com/equinor/radix-operator/blob/72214eba5cba486f5574d64f83b30e3037da7eb4/pipeline-runner/steps/preparepipeline/step.go#L197) should be [stored](https://github.com/equinor/radix-operator/blob/72214eba5cba486f5574d64f83b30e3037da7eb4/pipeline-runner/internal/runner/app.go#L143) by the pipeline runner in a new field ResolvedCommitID in [RadixJob.Status.PipelineRunStatus](https://github.com/equinor/radix-operator/blob/72214eba5cba486f5574d64f83b30e3037da7eb4/pkg/apis/radix/v1/radixjobtypes.go#L88).

DoD:
The resolved commit is stored in RadixJob.Status.PipelineRunStatus, and the user can see the commit for the job in Radix Web Console and in Radix CLI.

herda1:
- https://github.com/equinor/radix-operator/issues/1750

### DIFF
--- a/api-server/api/jobs/models/job_summary.go
+++ b/api-server/api/jobs/models/job_summary.go
@@ -1,6 +1,7 @@
 package models
 
 import (
+	"cmp"
 	"time"
 
 	"github.com/equinor/radix-common/utils/pointers"
@@ -155,13 +156,6 @@ type JobSummary struct {
 	// enum: branch,tag,""
 	// example: branch
 	GitRefType string `json:"gitRefType,omitempty"`
-
-	// ResolvedCommitID the commit ID resolved by the pipeline runner. Set when the pipeline job runs,
-	// even if CommitID was not provided in the pipeline arguments (e.g. manual trigger).
-	//
-	// required: false
-	// example: 4faca8595c5283a9d0f17a623b9255a0d9866a2e
-	ResolvedCommitID string `json:"resolvedCommitID,omitempty"`
 }
 
 // GetSummaryFromRadixJob Used to get job summary from a radix job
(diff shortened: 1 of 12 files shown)


api-server/api/jobs/models/job_summary.go
```diff
@@ -191,26 +192,33 @@ func GetSummaryFromRadixJob(job *radixv1.RadixJob) *JobSummary {
...
-		pipelineJob.CommitID = job.Spec.Build.CommitID
+		pipelineJob.CommitID = cmp.Or(GetResolvedCommitID(job), job.Spec.Build.CommitID)
 		pipelineJob.UseBuildKit = IsUsingBuildKit(job)
 		pipelineJob.UseBuildCache = IsUsingBuildCache(job)
 		pipelineJob.OverrideUseBuildCache = job.Spec.Build.OverrideUseBuildCache
 		pipelineJob.RefreshBuildCache = job.Spec.Build.RefreshBuildCache
 	case radixv1.Deploy:
 		pipelineJob.ImageTagNames = job.Spec.Deploy.ImageTagNames
-		pipelineJob.CommitID = job.Spec.Deploy.CommitID
+		pipelineJob.CommitID = cmp.Or(GetResolvedCommitID(job), job.Spec.Deploy.CommitID)
```
herda1: @Richard87 @nilsgstrabo 

Here, added 2 parameters on the assignment of commitID. 

The idea is:
-  if the pipeline has run/build - it will use the resolved commit id.
- if the pipeline has not run yet or its running - it will fall back to the commitId that the user has passed. 
If I don´t pass the job.Spec.Deploy.CommitID it will be empty until it is done. Should we leave it empty, or try to give an commitID while its processing as well 🤔

nilsgstrabo: I think is makes sense the way you have done it

Richard87: Thats pretty nice!

nilsgstrabo (approved):

PR #1845, opened 2026-07-14
Support in API to set and fetch annotations regarding federated credential migraton

jacobsolbergholm:
Once users have updated federated credentials for the new cluster issuer we want them to confirm that they have done so.
This will be set as an annotation on the RadixRegistration CR.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -247,10 +247,10 @@
 				"LOG_PRETTY": "true",
 				"OIDC_AZURE_ISSUER": "https://sts.windows.net/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/",
 				"OIDC_AZURE_AUDIENCE": "6dae42f8-4368-4678-94ff-3960e28e3630",
-				"OIDC_KUBERNETES_ISSUER": "https://northeurope.oic.prod-aks.azure.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/d759e0b0-6919-470c-82d9-848a3c0cacfd/",
-				"OIDC_KUBERNETES_AUDIENCE": "https://northeurope.oic.prod-aks.azure.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/d759e0b0-6919-470c-82d9-848a3c0cacfd/",
-				"RADIX_CLUSTERNAME": "weekly-26",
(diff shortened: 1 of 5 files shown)


.vscode/launch.json
```diff
@@ -247,10 +247,10 @@
 				"LOG_PRETTY": "true",
 				"OIDC_AZURE_ISSUER": "https://sts.windows.net/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/",
 				"OIDC_AZURE_AUDIENCE": "6dae42f8-4368-4678-94ff-3960e28e3630",
-				"OIDC_KUBERNETES_ISSUER": "https://northeurope.oic.prod-aks.azure.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/d759e0b0-6919-470c-82d9-848a3c0cacfd/",
-				"OIDC_KUBERNETES_AUDIENCE": "https://northeurope.oic.prod-aks.azure.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/d759e0b0-6919-470c-82d9-848a3c0cacfd/",
-				"RADIX_CLUSTERNAME": "weekly-26",
-				"CLUSTER_OIDC_ISSUERS": "https://northeurope.oic.prod-aks.azure.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/d759e0b0-6919-470c-82d9-848a3c0cacfd/,https://test-issuer.internal.radix.equinor.com",
+				"OIDC_KUBERNETES_ISSUER": "https://northeurope.oic.prod-aks.azure.com/3aa4a235-b6e2-48d5-9195-7fcf05b459b0/d9638477-2a54-40df-bb15-32f73505f631/",
```
herda1, line 250: which clientId was this?  🤔

herda1 (approved):

.vscode/launch.json
jacobsolbergholm: Good catch. Looking in the cluster, this is actually the issuer for weekly-28 and not 29. In this context it should not matter as it is for fetching the token based on federated credentials.

CLUSTER_OIDC_ISSUERS is for displaying/returning cluster issuers, so this would show the wrong issuers.
OIDC_KUBERNETES_ISSUER is used for authentication.

So for our purposes these should not matter, as long as they have matching federated credentials in Azure.

PR #1888, opened 2026-08-07
Return cron schedules for the job component.

### LINKED ISSUE #1885: Cron jobs - Return all schedules on job component level
A job component can have multiple schedules.
These should be returned through the API so they can be displayed in the web console.

jacobsolbergholm:


Richard87 (approved):
Looks good to me

PR #1900, opened 2026-08-12
fix: validate subdomain length constrain

### LINKED ISSUE #1519: Update webhook validation
Update validation to comply with IDNA2008.

Total length must be < 255

> "URI producers should use names that conform to the DNS syntax, even when use of DNS is not immediately apparent, and should limit these names to no more than 255 characters in length."
> 
> https://www.rfc-editor.org/rfc/rfc3986
> 
> "The DNS itself places only one restriction on the particular labels that can be used to identify resource records. That one restriction relates to the length of the label and the full name. The length of any one label is limited to between 1 and 63 octets. A full domain name is limited to 255 octets (including the separators)."
> 
> https://www.rfc-editor.org/rfc/rfc2181

https://stackoverflow.com/a/14402479

---
```
k get httproute -A -ojson --context c3-50 | jq '.items[] | .spec.hostnames[] |  split(".")[0] | length' -r | sort -h
```
The longest subdomain found is 61 characters, so blocking everything 64 or longer should be safe

herda1:
**Hostname label length validation:**

* Added a new error `ErrHostnameLabelTooLong` to `errors.go` to represent when a hostname label exceeds 63 characters.
* Implemented the `validateHostnameLabelLength` function to check that each label in a hostname is at most 63 characters, and integrated this check into the HTTPRoute validation logic in `validate_hr.go`. [[1]](diffhunk://#diff-0e9ed8a3b102341b594cd0ebe6139cb6f6e90d13dadaec8f52133716ae95349fR102-R114) [[2]](diffhunk://#diff-0e9ed8a3b102341b594cd0ebe6139cb6f6e90d13dadaec8f52133716ae95349fL83-R89)

**Testing:**

* Added unit tests to verify that hostnames with labels of exactly 63 characters are accepted, and those with labels of 64 characters are rejected, in `validate_hr_test.go`.

**Changes:** 
Screenshot 2026-08-13 at 14 15 36

### DIFF
--- a/webhook/validation/radixapplication/validate.go
+++ b/webhook/validation/radixapplication/validate.go
@@ -358,7 +358,8 @@ func envNameValidator(ctx context.Context, ra *radixv1.RadixApplication) ([]stri
 			return nil, []error{fmt.Errorf("environment %s: %w", env.Name, ErrInvalidEnvironmentNameLength)}
 		}
 		for _, component := range ra.Spec.Components {
-			if len(component.PublicPort) == 0 && component.Public == false {
+			//nolint:staticcheck // Public is supported for backwards compatibility.
+			if len(component.PublicPort) == 0 && !component.Public {
 				continue
 			}
 			namespace := utils.GetEnvironmentNamespace(ra.Name, env.Name)
--- a/webhook/validation/radixapplication/validate.go
+++ b/webhook/validation/radixapplication/validate.go
@@ -38,6 +38,7 @@ var (
 
 var (
 	offlineValidators = []validatorFunc{
+		applicationNameValidator,
 		deprecatedPublicUsageValidator,
 		componentJobNameValidator,
 		componentValidator,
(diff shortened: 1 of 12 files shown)


webhook/validation/radixapplication/validate.go
```diff
@@ -161,6 +162,15 @@ func branchNameValidator(ctx context.Context, ra *radixv1.RadixApplication) ([]s
 	return nil, nil
 }
 
+func applicationNameValidator(_ context.Context, ra *radixv1.RadixApplication) ([]string, []error) {
+	const maxApplicationNameLength = 63
```
nilsgstrabo: ra.Name must follow the same restrictions as RadixRegistration name, and can therefore be max 40 chars. Remember to update the errer text too

webhook/validation/radixapplication/validate.go
```diff
@@ -161,6 +162,15 @@ func branchNameValidator(ctx context.Context, ra *radixv1.RadixApplication) ([]s
 	return nil, nil
 }
```
nilsgstrabo, line 165: We should also find a way to test that the generated domain part, based on component, environment and application name, does not exceed 63 chars. We can do that by enumerating all environment names, and for each environment we enumerate through all public components, and generate the Radix specific domain name part, e.g. frontend-myapp-prod
There are currently no public methods for this. They are all private in /pkg/apis/deployments/gateway.go (function getActiveClusterHostName), but we can discuss how/where to make them publicly available to the validator code.

nilsgstrabo (approved):

PR #1934, opened 2026-08-26
fix(api-server): get gitRef and gitRefType from Annotations instead

### LINKED ISSUE #1871: api-server: resolve git-ref- and type from RD annotations
[GitRef and GitRefType](https://github.com/equinor/radix-operator/blob/d1cc6dc9959256ba823fc28eb84861661412eb22/api-server/api/deployments/models/deployment_builder.go#L265-L267) should be resolved from RadixDeployment annotations instead of the pipeline job that the deployment originates from.

The reason is that when a deployment is promoted to an environment, Radix Web Console does not show correct info of the origin (it is missing)

Example:
The deployment in `qa` is promoted from the deployment in `dev`, but since the deployment originates from a promote job, which does not have build info, the info in radix-web-console incorrectly says that is is built from `v*,` which is what is defined in radixconfig. It should instead say that it is built from branch main. 
Image

herda1:
- retrieved `gitRef` & `gitRefType` form Annotations
- moved and copied fields from `DeploymentSummaryPipelineJobInfo` into `DeploymentSummary` (since it was swagger ignored)
- changed the retrieval of the fields in former `DeploymentSummaryPipelineJobInfo` to be received from Annotations 
- removed `CommitID`  and `BuiltFromBranch` (as they are not used by radix-web-console)

### DIFF
--- a/api-server/api/deployments/models/deployment_builder.go
+++ b/api-server/api/deployments/models/deployment_builder.go
@@ -40,6 +40,8 @@ type deploymentBuilder struct {
 	repository         string
 	useBuildCache      *bool
 	refreshBuildCache  *bool
+	gitRef			  string
+	gitRefType		  string
 }
 
 // NewDeploymentBuilder Constructor for application deploymentBuilder
@@ -66,6 +68,8 @@ func (b *deploymentBuilder) WithRadixDeployment(rd *v1.RadixDeployment) Deployme
 		withActiveTo(activeTo).
 		withUseBuildCache(rd.Annotations[kube.RadixUseBuildCache]).
 		withRefreshBuildCache(rd.Annotations[kube.RadixRefreshBuildCache]).
+		withGitRef(rd.Annotations[kube.RadixGitRefAnnotation]).
+		withGitRefType(rd.Annotations[kube.RadixGitRefTypeAnnotation]).
 		WithGitCommitHash(rd.Annotations[kube.RadixCommitAnnotation]).
 		WithGitTags(rd.Annotations[kube.RadixGitTagsAnnotation])
 
@@ -93,6 +97,16 @@ func (b *deploymentBuilder) WithGitCommitHash(gitCommitHash string) DeploymentBu
 
 func (b *deploymentBuilder) WithGitTags(gitTags string) DeploymentBuilder {
 	b.gitTags = gitTags
+	return b
+}
+
+func (b *deploymentBuilder) withGitRef(gitRef string) *deploymentBuilder {
+	b.gitRef = gitRef
+	return b
+}
+
+func (b *deploymentBuilder) withGitRefType(gitRefType string) *deploymentBuilder {
+	b.gitRefType = gitRefType
 	return b
 }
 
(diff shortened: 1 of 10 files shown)


api-server/api/deployments/models/deployment_builder.go
```diff
@@ -261,8 +275,8 @@ func (b *deploymentBuilder) BuildDeployment() (*Deployment, error) {
 	}
 	if b.pipelineJob != nil {
 		deployment.BuiltFromBranch = b.pipelineJob.Spec.Build.Branch //nolint:staticcheck
```
herda1: rm

nilsgstrabo (dismissed):

api-server/api/models/environment_summary.go
```diff
@@ -40,7 +40,7 @@ func BuildEnvironmentSummaryList(rr *radixv1.RadixRegistration, ra *radixv1.Radi
 		deploymentSummary := getActiveDeploymentSummary(ra.GetName(), re.Spec.EnvName, rdList)
 		buildFromBranch := ""
 		if deploymentSummary != nil {
-			buildFromBranch = deploymentSummary.BuiltFromBranch
+			buildFromBranch = deploymentSummary.GitRef
```
herda1: @nilsgstrabo can you please verify this case for orphaned environments, whether it is correct this way?

herda1: Or I could do this instead: 
```
	if deploymentSummary != nil && deploymentSummary.GitRefType == "branch"{
			buildFromBranch = deploymentSummary.GitRef
	}
```

nilsgstrabo: It duplicates the functionality from the existing code, but now using a non-deprecated field. I think it is safer to remove it since the branch is no longer automatically built, and the current deployment can be a deployment promoted from another environment, and the original GitRef may not match the original branch name from the environment before it was removed from radixconfig.

I think it is better to remove the code. The environment will then be displayed with "Source (Deployed manually)" on the environment card in Radix Web Console, which is technically correct for orphaned environments.

herda1: Sounds right, I will remove it

nilsgstrabo (dismissed):

job-scheduler/pkg/watcher/watcher.go
```diff
@@ -120,11 +120,9 @@ func NewRadixBatchWatcher(ctx context.Context, radixClient radixclient.Interface
 }
 
 func notify(ctx context.Context, notifier notifications.Notifier, ev events.Event, newRadixBatch *radixv1.RadixBatch, updatedJobStatuses []radixv1.RadixBatchJobStatus) {
-	go func() {
-		if err := notifier.Notify(ev, newRadixBatch, updatedJobStatuses); err != nil {
-			log.Ctx(ctx).Error().Err(err).Msg("failed to notify")
-		}
-	}()
+	if err := notifier.Notify(ev, newRadixBatch, updatedJobStatuses); err != nil {
```
nilsgstrabo: I think we have to revert this back to running it in a go-routine. The notifier sends a http request to an endpoint configured for the job in radixconfig, and if the endpoint does not respond, or is slow, then it will block other events in the eventhandler. Our current solution with using a go-routines is not good since we potentially end up with a lot of active go-routines if the endpoint used by the notifier is unresponsive. We should probably consider refactoring, for example by putting the objects on a queue, and then have a loop processing items from the queue, similar to how the controllers in the operator works.

herda1: reverted it back, creating a new issue about the flaky test

nilsgstrabo (approved):

PR #1938, opened 2026-08-31
fix(api-server): remove deprecated secret type 'azure-blob-fuse-volume'

### LINKED ISSUE #1937: Remove deprecated secret type from api-server
Secret type [SecretTypeAzureBlobFuseVolume](https://github.com/equinor/radix-operator/blob/e1149a709b4925b4878a3934bae8dc2976492422/api-server/api/secrets/models/secret.go#L69) is no longer used. Was most likely used by the first version of the blob driver.

nilsgstrabo:


Richard87 (approved):
Godkjent

